PR #1, opened 2022-06-20
cli: add basic configuration commands

audrium:
Added basic configuration commands: `version`, `ping`, `info`

To test:
- `go run .`
- `go run . ping`
- `go run . info --json`

### DIFF
--- a/validation/utils.go
+++ b/validation/utils.go
@@ -1,0 +1,29 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package validation
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+func ValidateAccessToken(token string) {
+	if strings.TrimSpace(token) == "" {
+		fmt.Println("Error: Please provide your access token by using the -t/--access-token flag, or by setting the REANA_ACCESS_TOKEN environment variable.")
+		os.Exit(1)
+	}
+}
+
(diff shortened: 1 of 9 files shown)


validation/utils.go
```diff
@@ -0,0 +1,27 @@
...
+
+package validation
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+func ValidateTokenAndServerURL(token string, serverURL string) {
```
tiborsimko, line 17: Will it make sense to validate each separately, i.e. to have two functions?

(Otherwise we may rename the function, because "validate token and server url" may read ambiguously. Perhaps `validateServerUrlAndAccessToken()`.)

validation/utils.go
```diff
@@ -0,0 +1,27 @@
...
+	if strings.TrimSpace(token) == "" {
+		fmt.Println("Error: Please provide your access token by using the -t/--access-token flag, or by setting the REANA_ACCESS_TOKEN environment variable.")
+		os.Exit(1)
+	}
+
+	if strings.TrimSpace(serverURL) == "" {
+		fmt.Println("Please set REANA_SERVER_URL environment variable.")
```
tiborsimko, line 24: The first validation failure message has leading "Error:", the second validation failure message does not.  Can we harmonise?  

(BTW in the future we may want to introduce some nice info/debug/error logging library with colouring and whatnot.)

validation/utils.go
audrium: yeah, breaking it into separate functions make sense, especially since serverURL could later be validated in the swagger client before calling the api

tiborsimko (approved):

PR #6, opened 2022-06-21
list: initial "bare" partial implementation

tiborsimko:
Adds initial "bare" partial implementation of the `list` command that
was created mostly for performance testing back in January 2022. Example
of results: listing O(100) workflows took 0.2 sec in Go, 1.5 sec in
Python, and 2.1 sec in Python AppImage.

We can use this "bare" version to run more performance tests (and to
compare to the swagger version that is coming soon.)

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -1,0 +1,126 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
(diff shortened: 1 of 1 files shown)


audrium (approved):
LGTM!

PR #7, opened 2022-06-21
cli: add swagger client and filter option to list command

audrium:
- Added automatically generated API client (https://github.com/go-swagger/go-swagger)
- Generated API client files with the updated openapi specification (https://github.com/reanahub/reana-commons/pull/355)
- Updated all available commands to use swagger api client
- Added `filter` option to the `list` command

To test:
- `go mod tidy`
- check if `ping` and `info` commands still works as expected
- try to play with `list` command `filter` option. e.g. `go run . list --filter name=roofit --filter status=running`


tiborsimko (approved):
Works nice and keeps the speed, e.g. for O(800):

```console
$ hyperfine './reana-client-go-old-list list >& /dev/null' './reana-client-go list >& /dev/null'
Benchmark 1: ./reana-client-go-old-list list >& /dev/null
  Time (mean ± σ):     758.1 ms ±  39.2 ms    [User: 31.8 ms, System: 17.9 ms]
  Range (min … max):   708.7 ms … 811.6 ms    10 runs

Benchmark 2: ./reana-client-go list >& /dev/null
  Time (mean ± σ):     746.7 ms ±  67.2 ms    [User: 35.3 ms, System: 19.1 ms]
  Range (min … max):   686.1 ms … 921.7 ms    10 runs

Summary
  './reana-client-go list >& /dev/null' ran
    1.02 ± 0.11 times faster than './reana-client-go-old-list list >& /dev/null'
```

PR #13, opened 2022-07-06
cli: Hide access token from the command description

### LINKED ISSUE #11: cli: hide access token from the command description
Right now when one checks a command description, access token is being displayed. It happens because Cobra library displays the default value used for command. We should hide this sensitive information, but still set the value from environmental variable by default.

Example:
```
$ go run . ping --help

Check connection to REANA server.
...
Flags:
  -t, --access-token string   Access token of the current user. (default "XXX")
```

BrunoRosendo:
closes #11

Hides the access token from the command description when using `--help`. Did it by having our own logic for a default value, instead of using Cobra's feature

### DIFF
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -40,6 +40,9 @@ func newInfoCmd() *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			jsonOutput, _ := cmd.Flags().GetBool("json")
 			token, _ := cmd.Flags().GetString("access-token")
+			if token == "" {
+				token = os.Getenv("REANA_ACCESS_TOKEN")
+			}
 			serverURL := os.Getenv("REANA_SERVER_URL")
 			validation.ValidateAccessToken(token)
 			validation.ValidateServerURL(serverURL)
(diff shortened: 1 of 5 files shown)


cmd/info.go
```diff
@@ -40,6 +40,10 @@ func newInfoCmd() *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			jsonOutput, _ := cmd.Flags().GetBool("json")
 			token, _ := cmd.Flags().GetString("access-token")
+			if token == "" {
+				token = os.Getenv("REANA_ACCESS_TOKEN")
+			}
+
```
audrium, line 46: Minor: do we need these empty lines after this check? IMO it logically separates a `token` from `serverURL` which are kind of related, WDYT?

audrium (commented):
Congrats with your first contribution!

I left one cosmetic comment, but it works nicely!
Could you please rebase your commit on top of latest changes and add yourself
as an author [here](https://github.com/reanahub/reana-client-go/blob/master/AUTHORS.rst)?

Also, can you add an empty line between your commit and closes statement like described in the [development wiki](https://github.com/reanahub/reana/wiki/Developing#structure-and-style-of-commits)? P.S. usually we keep the commit messages in lowercase, it would be nice to stick to this convention

cmd/info.go
BrunoRosendo: Sure, I just thought it was too bloated but maybe it makes sense to put them together, I can just that

audrium (approved):
Nicely done

PR #22, opened 2022-07-07
cli: fix reana server url being verified when not used

### LINKED ISSUE #12: cli: REANA server URL validation is done also for commands which do not use it
If a user did not set an environment variable for REANA server and tries to check a `version` command (or any other that does not require `REANA_SERVER_URL`) it will fail like this:

```
$ go run . version
Error: Environment variable REANA_SERVER_URL is not set
exit status 1
```

It happens because we validate it in our [apiClient](https://github.com/reanahub/reana-client-go/blob/master/client/client.go#L31) which is triggered on init. We should validate if this value is present only for commands which uses `apiClient`.

BrunoRosendo:
closes #12

Prevents REANA server URL from being validated when not used, by transforming ApiClient into a function, so that it is only initialized when a command actually needs it

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -12,7 +12,14 @@ import (
 	httptransport "github.com/go-openapi/runtime/client"
 )
 
-var ApiClient = newApiClient()
+var apiClient *API
+
+func ApiClient() *API {
+	if apiClient == nil {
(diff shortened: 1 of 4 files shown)


audrium (approved):
Works nicely!

PR #25, opened 2022-07-07
GitHub actions setup

### LINKED ISSUE #4: ci: add GitHub Actions support
Integrate GitHub Actions platform support for continuous integration like we have in all other repos ([reana-client example](https://github.com/reanahub/reana-client/blob/master/.github/workflows/ci.yml)) which would run the test suit and perform all other necessary checks

BrunoRosendo:
closes #4 

Adds CI support, featuring go tests and golangci-lint

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,14 +48,11 @@ jobs:
               with:
                   go-version: '>=1.18.X'
 
-            - name: Install golines
-              run: go install github.com/segmentio/golines@latest
-
             - name: Run golines
               run: |
-                  if [[ $(golines --dry-run ./) ]]; then
+                  if [[ $(make golines | wc -l) -gt 2 ]]; then
                       echo "Golines would make the following changes:"
-                      golines --dry-run ./
+                      make golines
                       exit 1
                   else
                       echo "No file exceeds the line size limit"
--- a/Makefile
+++ b/Makefile
@@ -41,4 +41,12 @@ update: # Update go module dependencies.
 	go get -u
 	go mod tidy
 
-.PHONY: all build clean help release swagger-generate-client swagger-validate-specs test update
+lint: # Run lint checks
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+	golangci-lint run --enable=gofmt
+
+golines: # Run line size checks
+	go install github.com/segmentio/golines@latest
+	golines --dry-run ./
+
+.PHONY: all build clean help release swagger-generate-client swagger-validate-specs test update lint golines


BrunoRosendo:
It's in draft until #15 is merged, so that we can use it in the github action

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Setup Go
+              uses: actions/setup-go@v3
+              with:
+                  go-version: '>=1.18.X'
+            - run: go test ./... # TODO Replace with make test after #15
+
+    golangci:
```
audrium, line 23: Minor: maybe we should call it `go-lint` to be more in line with how we name other ci jobs?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+        runs-on: ubuntu-20.04
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Setup Go
+              uses: actions/setup-go@v3
+              with:
+                  go-version: '>=1.18.X'
+
+            - name: golangci-lint
```
audrium, line 33: Very minor: would be good to stick to convention and to capitalise the step name

audrium (commented):
Good work!
Left few cosmetic comments, but it looks good.
I think the only thing which is missing is to check the formatting of the code. We could use something like  `go fmt -n` to print the files that would be formatted or try to look for any GA for that.

P.S. This PR needs a rebase.

BrunoRosendo:
@audrium golangci also supports gofmt (see https://golangci-lint.run/usage/linters/). Maybe I could try to configure it in order to run gofmt

audrium:
> @audrium golangci also supports gofmt (see https://golangci-lint.run/usage/linters/). Maybe I could try to configure it in order to run gofmt

Indeed! Nice, I think adding `--enable=gofmt` should do it

BrunoRosendo:
@audrium Any others you might find interesting to enable? I only found wsl to be interesting (Whitespace Linter - Forces you to use empty lines). What do you think?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+            - name: Setup Go
+              uses: actions/setup-go@v3
+              with:
+                  go-version: '>=1.18.X'
+
+            - name: Install golines
+              run: go install github.com/segmentio/golines@latest
+
+            - name: Run golines
+              run: golines .
```
audrium, line 55: I think we should use `--dry-run ` option here, because by default it just prints everything

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+              with:
+                  go-version: '>=1.18.X'
+
+            - name: Install golines
+              run: go install github.com/segmentio/golines@latest
+
+            - name: Run golines
+              run: |
+                  if [[ $(golines --dry-run ./) ]]; then
+                      echo "Golines would make the following changes:"
```
tiborsimko, line 57: Please make this test part of `make test`, then you would be able to use `make test` here, and people could run the same test locally, to catch formatting troubles early, before we consume GA cycles.

audrium: I think it should still be nice to have this check in the `Makefile` as well. Perhaps in the `Makefile`s target you can detect the line count and do `exit 1` there? That should do the trick I think

BrunoRosendo: @audrium Do you mean it should be part of the `test` target? Or keep the `golines` target and do `exit 1` there?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+        runs-on: ubuntu-20.04
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Setup Go
+              uses: actions/setup-go@v3
+              with:
+                  go-version: '>=1.18.X'
+
+            - name: Install golines
```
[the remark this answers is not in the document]
BrunoRosendo: Oh but I don't call the makefile here. Did you see [this comment](https://github.com/reanahub/reana-client-go/pull/25/files#r916860919)?

audrium (approved):
LGTM!

PR #26, opened 2022-07-08
setup: new make clean, help, update targets

tiborsimko:
Introduces new make targets clean, help, update. Renames OpenAPI-related targets for better consistency.


audrium (approved):
LGTM!

PR #33, opened 2022-07-11
cli: implement du command

### LINKED ISSUE #18: cli: implement `du` command
Create a new `du` command that replicates all the functionality we have in `reana-client`.
It should also include `summarize` and `human-readable` flags and return exactly the same output as `reana-client du`.

Example:
```
$ reana-client du -w workflow -h
SIZE      NAME
152 KiB   ./results/data.root
16 KiB    ./results/plot.png
172 KiB   ./results
4 KiB     ./code/gendata.C
4 KiB     ./code/fitdata.C
12 KiB    ./code
188 KiB   .
```

BrunoRosendo:
closes #18


cmd/du.go
```diff
@@ -0,0 +1,116 @@
...
+			validation.ValidateServerURL(serverURL)
+			validation.ValidateWorkflow(workflow)
+			du(cmd)
+		},
+	}
+
+	cmd.Flags().StringP("access-token", "t", "", "Access token of the current user.")
+	cmd.Flags().StringP("workflow", "w", os.Getenv("REANA_WORKON"), "List all runs of the given workflow.")
+	cmd.Flags().BoolP("summarize", "s", false, "Display total.")
+	cmd.Flags().BoolP("human-readable", "r", false, "Show disk size in human readable format.")
```
BrunoRosendo, line 59: `-h` does not work here because it is already used for `--help`

cmd/du.go
```diff
@@ -0,0 +1,130 @@
...
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"os"
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/utils"
+	"reanahub/reana-client-go/validation"
```
audrium, line 18: When I saved the file for me the imports where automatically sorted like this:
```
	"fmt"
	"os"
	"reanahub/reana-client-go/client"
	"reanahub/reana-client-go/client/operations"
	"reanahub/reana-client-go/utils"
	"reanahub/reana-client-go/validation"

	"github.com/spf13/cobra"
```
Not sure if it's some kind of external plugin which comes with go extension in VS Code, or what. We might need to figure this out in the future and align these kind of things to make it consistent in all the files

audrium (approved):
Nice job! It woks well!

I left one minor comment for the column width. I also think we should create an issue for `summarize` flag, since it's not working now. So after [this is fixed](https://github.com/reanahub/reana-server/issues/521) we should not forget to implement it

BrunoRosendo:
@audrium I created [this issue](https://github.com/reanahub/reana-server/issues/521) for the summarize problem

cmd/du.go
BrunoRosendo: I think I found the problem, it's `goimports` that was probably activated for you and not for me. We can include it in the `golangci-lint` to make it consistent for everyone

BrunoRosendo: I added it to the lint checks. See if you like it

audrium: Yeah, looks nice, thanks for a quick action!

audrium:
> @audrium I created [this issue](https://github.com/reanahub/reana-server/issues/521) for the summarize problem

Yes, I saw it and I linked it to my previous comment. My point was that after fixing the issue in `reana-server` side we should not forget to implement this flag properly here in `du` command. It's better to create an issue for that rather than to keep it in mind

PR #34, opened 2022-07-12
cli: common and better way to print tables

### LINKED ISSUE #21: cli: investigate a better way to print tables
Lots of our commands require printing tabular information. Right now we do it manually [like this](https://github.com/reanahub/reana-client-go/blob/master/cmd/list.go#L123-L131). It does the job well, but the problem is that the column widths are hardcoded and static. It does not work well when information exceeds predefined column width.

Example:
```
$ go run . list
NAME                                   RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS
loooooooooooooooooooooooooooooong-workflow-name 1            2022-07-07T12:40:15   2022-07-07T12:40:26   -                     running
workflow                               7            2022-07-06T07:40:23   2022-07-06T07:40:34   2022-07-06T07:40:50   finished
``` 

It would be nice to find a nice library which could handle all these kind of printing issues for us. After quick search I found some candidates: 
- https://github.com/jedib0t/go-pretty
- https://github.com/olekukonko/tablewriter

This issue consists of two parts:
- investigate and pick the best library for our needs (or maybe we don't need an external library and there is a better way to do it without it?)
- implement the changes needed for using this library in all the commands

BrunoRosendo:
closes #21

After considering some package options (including go-pretty, tablewriter, table and tabwriter), I reached the conclusion that `go-pretty` would be better, for the following reasons:
- Many users (only falling behind tablewriter)
- Simple way to create and customize tables
- Provides some predefined styles
- Also provides ways to customize lists, progress bars and text, which could prove useful later
- Active developers (unlike tablewriter, whose last update was more than a year ago)

Right now, the tables follow a very simplistic style:
image

However, there are also coloured and more complex styles like this one:
image

Which kind of style do you think we should choose here?


BrunoRosendo:
> Looks nice! Is it possible to print the table without any borders just like we currently do in `reana-client`? I quickly checked that Helm project uses [uitable](https://github.com/gosuri/uitable). Here is and example: [helm list command](https://github.com/helm/helm/blob/823d92942119dfa07fd763bf6306bd67686811a8/cmd/helm/list.go#L179-L186)

Yes, it's possible. Here are different options:
image
image
image

As for the uitable, do you think it has advantes over go-pretty?

audrium:
> Yes, it's possible. Here are different options:

Looks very nice! I guess we could start with the most simple design and implement the tables without any borders as we currently do in python client. In the future we could customize and tweak the appearance if we want. It's good that this library provides all these options!

> As for the uitable, do you think it has advantes over go-pretty?

No, it looks like outdated project already, last commit was in 2019, but I've added to document what others are using. So I personally think we can proceed with `go-pretty`, it has all these advantages you listed in the top compared to others

audrium (approved):
Nice work!

Issue #27, opened 2022-07-08
cli: implement all `list` command flags


audrium:
Right now we have a basic version of `list` command with some flags already working. We need to implement all the other flags and replicate all the functionality we have in `reana-client`.

Here is a list of all the flags `list` command have:
- [x] `-w, --workflow TEXT List all runs of the given workflow.`
- [x] `-s, --sessions List all open interactive sessions.`
- [ ] `--format TEXT Format output according to column titles or column values. `
- [x]  `--json Get output in JSON format.`
- [x] `--all Show all workflows including deleted ones.`
- [x] `-v, --verbose Print out extra information: workflow id, user id, disk usage, progress, duration.`
- [x] `-h, --human-readable Show disk size in human readable format.`
- [ ] `--sort TEXT Sort the output by specified column`
- [x] `--filter TEXT Filter workflow that contains certain filtering criteria.`
- [x] `--include-duration Include the duration of the workflows in seconds.`
- [x] `--include-progress Include progress information of the workflows.`
- [x] `--include-workspace-size Include size information of the workspace.`
- [x] `--show-deleted-runs Include deleted workflows in the output.`
- [x] pagination (`--page`, `--size`)

This is an epic issue which can be used to link all the PRs used to develop these features

Issue #16, opened 2022-07-07
cli: improve api client error handling


audrium:
Currently in case of any errors we simply [print the error message](https://github.com/reanahub/reana-client-go/blob/master/cmd/ping.go#L54) to the user. We should improve the user experience by handling basic error cases to return more understandable messages for the user.

Example if someone tries to use a wrong access token:
```
$ go run . ping -t "test"
Error:  [GET /api/you][403] getYouForbidden
exit status 1
```
This message might be confusing to the users. Inspiration of how we handle these cases for the `ping` command in `reana-client` could be found [here](https://github.com/reanahub/reana-client/blob/e06517f1d0119031c8c7bbbbf7511b2eea56c8e6/reana_client/api/client.py#L51-L62).

PR #35, opened 2022-07-14
cli: implement more flags in the list command

BrunoRosendo:
Contributes to #27 

Introduces all the flags functionality, except format, sort and json. Some other features, such as the session headers, were left with a todo mark.
This seemed like a good mark to ask for a review. The following work will be to fix all the TODOs and implement the flags that are still missing

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -15,7 +15,7 @@ import (
 	"reanahub/reana-client-go/client/operations"
 	"reanahub/reana-client-go/utils"
 	"reanahub/reana-client-go/validation"
-	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 )
@@ -68,10 +68,24 @@ func newListCmd() *cobra.Command {
 
 	cmd.Flags().StringP("access-token", "t", "", "Access token of the current user.")
 	cmd.Flags().StringP("workflow", "w", "", "List all runs of the given workflow.")
-	cmd.Flags().StringP("sessions", "s", "", "List all open interactive sessions.")
-	cmd.Flags().String("format", "", listFormatFlagDesc)
-	cmd.Flags().BoolP("json", "", false, "Get output in JSON format.")
+	cmd.Flags().BoolP("sessions", "s", false, "List all open interactive sessions.")
+	cmd.Flags().StringSlice("format", []string{}, listFormatFlagDesc)
+	cmd.Flags().Bool("json", false, "Get output in JSON format.")
+	cmd.Flags().Bool("all", false, "Show all workflows including deleted ones.")
+	cmd.Flags().
+		BoolP("verbose", "v", false, `Print out extra information: workflow id, user id, disk usage,
+progress, duration.`)
+	cmd.Flags().BoolP("human-readable", "r", false, "Show disk size in human readable format.")
+	cmd.Flags().String("sort", "CREATED", "Sort the output by specified column.")
 	cmd.Flags().StringArray("filter", []string{}, listFilterFlagDesc)
+	cmd.Flags().
+		Bool("include-duration", false, `Include the duration of the workflows in seconds. In case a workflow is in
+progress, its duration as of now will be shown.`)
+	cmd.Flags().Bool("include-progress", false, "Include progress information of the workflows.")
+	cmd.Flags().Bool("include-workspace-size", false, "Include size information of the workspace.")
+	cmd.Flags().Bool("show-deleted-runs", false, "Include deleted workflows in the output.")
+	cmd.Flags().Int64("page", 1, "Results page number (to be used with --size).")
+	cmd.Flags().Int64("size", 0, "Size of results per page (to be used with --page).")
 
 	return cmd
 }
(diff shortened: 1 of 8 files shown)


cmd/list.go
```diff
@@ -82,57 +98,141 @@ func list(cmd *cobra.Command) {
...
+			case "user":
+				value = workflow.User
+			case "size":
+				if humanReadable {
+					value = workflow.Size.HumanReadable
+				} else {
+					value = workflow.Size.Raw
+				}
+			case "progress":
+				// TODO check if I have to be careful with nils
```
BrunoRosendo, line 183: Do you know if it's possible for this information to be missing? The Python version checked for nulls but I'm not sure and it'll make the code more confusing

BrunoRosendo: I noticed it actually comes as 0 (GO's default value) so I'll try to manage something

cmd/list.go
```diff
@@ -82,62 +98,198 @@ func list(cmd *cobra.Command) {
...
-		rows = append(rows, row)
-	}
-
-	utils.DisplayTable(header, rows)
+func getWorkflowNameAndRunNumber(workflowName string) (string, string) {
+	workflowNameAndRunNumber := strings.SplitN(workflowName, ".", 2)
+	if len(workflowNameAndRunNumber) < 2 {
+		return workflowName, ""
+	}
+	return workflowNameAndRunNumber[0], workflowNameAndRunNumber[1]
```
mdonadoni, line 244: We will probably need this in other commands, should we already move it to `utils`?

cmd/list.go
```diff
@@ -82,62 +98,198 @@ func list(cmd *cobra.Command) {
...
 
-func displayOptionalField(value *string) string {
+func buildListHeaders(
+	runType string,
+	verbose, includeWorkspaceSize, includeProgress, includeDuration bool,
+) map[string][]any {
+	verboseHeaders := []any{"id", "user"}
+	workspaceSizeHeader := []any{"size"}
+	progressHeader := []any{"progress"}
+	durationHeader := []any{"duration"}
```
mdonadoni, line 254: What do you think about removing these? We can then just do it like this:
```go
if verbose || includeProgress {
	headers[runType] = append(headers[runType], "progress")
}
```

cmd/list.go
```diff
@@ -82,62 +98,198 @@ func list(cmd *cobra.Command) {
...
+		return workflowName, ""
+	}
+	return workflowNameAndRunNumber[0], workflowNameAndRunNumber[1]
 }
 
-func displayOptionalField(value *string) string {
+func buildListHeaders(
+	runType string,
+	verbose, includeWorkspaceSize, includeProgress, includeDuration bool,
+) map[string][]any {
```
mdonadoni, line 250: What do you think about returning only the `[]string` headers, instead of the map? The type is already known, so we do not need to return the headers for both `batch` and `interactive`.

PS: I think it is better to use `string` instead of `any` here!

cmd/list.go
```diff
@@ -82,62 +98,198 @@ func list(cmd *cobra.Command) {
...
+			case "size":
+				if humanReadable {
+					value = workflow.Size.HumanReadable
+				} else {
+					value = workflow.Size.Raw
+				}
+			case "progress":
+				progress := workflow.Progress
+				finishedInfo := displayOptionalFloatField(progress.Finished.Total)
+				totalInfo := displayOptionalFloatField(progress.Total.Total)
```
mdonadoni, line 185: We should check why these values are passed as floats instead of integers, given that they should be the number of jobs running/finished.

cmd/list.go
```diff
@@ -82,62 +98,198 @@ func list(cmd *cobra.Command) {
...
+		headers[runType] = append(headers[runType], progressHeader...)
+	}
+	if verbose || includeDuration {
+		headers[runType] = append(headers[runType], durationHeader...)
+	}
+
+	return headers
+}
+
+func displayOptionalStringField(value *string) string {
```
mdonadoni, line 283: As before, this function does not display the value passed, so we should probably not call it `display...`. WDYT?

cmd/list.go
```diff
@@ -82,62 +98,198 @@ func list(cmd *cobra.Command) {
...
 	if value == nil {
 		return "-"
 	}
 	return *value
 }
+
+func displayOptionalFloatField(value float64) string {
```
mdonadoni, line 290: As before, this function does not display the value passed, so we should probably not call it `display...`. WDYT?

cmd/list.go
```diff
@@ -68,10 +69,25 @@ func newListCmd() *cobra.Command {
...
+	cmd.Flags().String("sort", "CREATED", "Sort the output by specified column.")
 	cmd.Flags().StringArray("filter", []string{}, listFilterFlagDesc)
+	cmd.Flags().
+		Bool("include-duration", false, `Include the duration of the workflows in seconds. In case a workflow is in
+progress, its duration as of now will be shown.`)
+	cmd.Flags().Bool("include-progress", false, "Include progress information of the workflows.")
+	cmd.Flags().Bool("include-workspace-size", false, "Include size information of the workspace.")
+	cmd.Flags().Bool("show-deleted-runs", false, "Include deleted workflows in the output.")
+	cmd.Flags().Int64("page", 1, "Results page number (to be used with --size).")
+	// TODO Python used null but it doesn't work here, should we use 0 or something else?
```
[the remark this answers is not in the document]
BrunoRosendo: I think that makes sense, I'll do option (2)

cmd/list.go
```diff
@@ -82,62 +98,198 @@ func list(cmd *cobra.Command) {
...
+			switch col {
+			case "id":
+				value = workflow.ID
+			case "user":
+				value = workflow.User
+			case "size":
+				if humanReadable {
+					value = workflow.Size.HumanReadable
+				} else {
+					value = workflow.Size.Raw
```
mdonadoni, line 180: We should check why this field is returned as a float, given that it should be the (integer) number of bytes. Also, given that it is a float, it is sometimes displayed in exponential form (e.g. `1.359872e+06`)

cmd/list.go
```diff
@@ -68,10 +69,25 @@ func newListCmd() *cobra.Command {
 
 	cmd.Flags().StringP("access-token", "t", "", "Access token of the current user.")
 	cmd.Flags().StringP("workflow", "w", "", "List all runs of the given workflow.")
-	cmd.Flags().StringP("sessions", "s", "", "List all open interactive sessions.")
-	cmd.Flags().String("format", "", listFormatFlagDesc)
+	cmd.Flags().BoolP("sessions", "s", false, "List all open interactive sessions.")
+	cmd.Flags().StringArray("format", []string{}, listFormatFlagDesc)
 	cmd.Flags().BoolP("json", "", false, "Get output in JSON format.")
+	cmd.Flags().BoolP("all", "", false, "Show all workflows including deleted ones.")
```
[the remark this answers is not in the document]
BrunoRosendo: Definitely, I did not notice that

cmd/list.go
```diff
@@ -82,57 +98,141 @@ func list(cmd *cobra.Command) {
...
+	listParams.SetVerbose(&verbose)
+	listParams.SetPage(&page)
 	listParams.SetWorkflowIDOrName(&workflow)
 	listParams.SetStatus(statusFilters)
 	listParams.SetSearch(&searchFilter)
+	listParams.SetIncludeProgress(&includeProgress)
+	listParams.SetIncludeWorkspaceSize(&includeWorkspaceSize)
+	if size > 0 {
+		listParams.SetSize(&size)
+	}
```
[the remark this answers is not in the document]
BrunoRosendo: Yes, that will work well after using the changed() from the comment above

cmd/list.go
```diff
@@ -82,57 +98,141 @@ func list(cmd *cobra.Command) {
...
+		includeWorkspaceSize,
+		includeProgress,
+		includeDuration,
+	)
+	displayListPayload(listResp.Payload, headers[runType], jsonOutput, humanReadable)
+}
+
+func displayListPayload(
+	p *operations.GetWorkflowsOKBody,
+	header []any,
```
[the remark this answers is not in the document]
BrunoRosendo: I don't really like using `any` either, but using `string` has a problem:
- The library used to draw tables had a type `Row`, that's basically a new name to `[]any`
- To not depend in the framework and use it with every type, I made the `utils.DisplayTable` to receive the type `any`
- Unfortunately, GO doesn't let me pass a `string` type when it is expecting `any`, hence why the headers all have `[]any` instead of `[]string`
Do you know a better way to do this?

cmd/list.go
BrunoRosendo: Makes sense. As for the `any`, it's the same problem as the comment above

cmd/list.go
BrunoRosendo: Do we just need to update the openapi or is it something server sided?

cmd/list.go
BrunoRosendo: I think updating the api worked

cmd/list.go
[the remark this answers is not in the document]
BrunoRosendo: Hmm I see, I think it's worth making the change then

utils/display.go
```diff
@@ -9,23 +9,41 @@ under the terms of the MIT License; see LICENSE file for more details.
...
 	t.Style().Options.SeparateHeader = false
 	t.Render()
 }
+
+func DisplayJsonOutput(output any) {
+	byteArray, err := json.MarshalIndent(output, "", "  ")
+
+	if err != nil {
+		fmt.Println("Error: ", err)
+	}
```
mdonadoni, line 46: We should either call `os.Exit(1)` or return the error and handle it in `displayListPayload`. What do yoy think?

BrunoRosendo: I think we should call `os.Exit(1)` since `displayListPayload` (and the other functions) should always call this with the right arguments. It's also what's being done in other places, #16 is meant to improve this behaviour everywhere

mdonadoni (approved):
Works nicely!

PS: The client needs to be regenerated, but we can do that in the next PR that implements the rest of the flags.

PR #39, opened 2022-07-19
cli: sort and format in list command

### LINKED ISSUE #27: cli: implement all `list` command flags

BrunoRosendo:
closes #27

Implement the remaining flags in the `list` command: `sort` and `format`

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -15,7 +15,11 @@ import (
 	"reanahub/reana-client-go/client/operations"
 	"reanahub/reana-client-go/utils"
 	"reanahub/reana-client-go/validation"
+	"sort"
+	"strings"
 	"time"
+
+	"golang.org/x/exp/slices"
 
 	"github.com/spf13/cobra"
 )
@@ -101,6 +105,7 @@ func list(cmd *cobra.Command, serverURL string) {
 	showAll, _ := cmd.Flags().GetBool("all")
 	verbose, _ := cmd.Flags().GetBool("verbose")
 	humanReadable, _ := cmd.Flags().GetBool("human-readable")
+	sortColumn, _ := cmd.Flags().GetString("sort")
 	filters, _ := cmd.Flags().GetStringArray("filter")
 	includeDuration, _ := cmd.Flags().GetBool("include-duration")
 	includeProgress, _ := cmd.Flags().GetBool("include-progress")
@@ -150,12 +155,12 @@ func list(cmd *cobra.Command, serverURL string) {
 		includeProgress,
 		includeDuration,
 	)
-	displayListPayload(listResp.Payload, serverURL, token, header, jsonOutput, humanReadable)
+	displayListPayload(listResp.Payload, serverURL, token, sortColumn, header, jsonOutput, humanReadable)
 }
 
 func displayListPayload(
 	p *operations.GetWorkflowsOKBody,
-	serverURL, token string,
+	serverURL, token, sortColumn string,
 	header []string,
 	jsonOutput, humanReadable bool,
 ) {
@@ -213,6 +218,8 @@ func displayListPayload(
 		}
 		data = append(data, row)
 	}
+
+	sortListData(data, header, sortColumn)
 
 	if jsonOutput {
 		jsonData := make([]map[string]any, len(data))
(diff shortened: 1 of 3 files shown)


cmd/list.go
```diff
@@ -289,3 +310,36 @@ func getOptionalIntField(value int64) string {
...
+		switch value1.(type) {
+		case int64:
+			return value1.(int64) > value2.(int64)
+		case float64:
+			return value1.(float64) > value2.(float64)
+		case string:
+			return value1.(string) > value2.(string)
+		default:
+			return false // Keep order if any other type
+		}
```
BrunoRosendo, line 343: I tried to look for a general way to convert the values into an ordered interface but I don't think GO allows that. So I just included in the switch case the types used in the list command. Should I include others? (there are a lot of them, e.g., int, int32, float32, uint8, etc.)

BrunoRosendo: I could also do the fmt.Sprint() for the default value, instead of keeping the order

utils/utils.go
```diff
@@ -126,6 +126,52 @@ func ParseListFilters(filter []string, filterNames []string) ([]string, string)
...
+		if len(filterNameAndValue) < 2 {
+			parsedFilters[filterNameAndValue[0]] = nil
+		} else {
+			parsedFilters[filterNameAndValue[0]] = filterNameAndValue[1]
+		}
+	}
+	return parsedFilters
+}
+
+func FormatData(data *[][]any, header *[]string, formatFilters map[string]any) {
```
mdonadoni, line 142: I think we can make this function a little bit more readable and compact:

1. Use `row` instead of `j` and `col` instead of `i`
3. Chaining the if/else so that we do not use `continue`

Something similar to this:
```go
func FormatData(data *[][]any, header *[]string, formatFilters map[string]any) {
	if len(formatFilters) == 0 {
		return
	}

	for col := 0; col < len(*header); col++ {
		filter, include := formatFilters[(*header)[col]]

		if !include {
			// Remove the column from header and data
			...
		} else if filter != nil {
			// Remove rows not containing filter
			...
		}
	}
}
```
What do you think?

PS: note that however it is usually discouraged to modify list/array/slice while iterating over them, as it can cause many subtle bugs

mdonadoni: There is also another issue: when passing multiple format filters, such as `--format run_number,name`, the order in which the column headers are passed is not respected.
```console
$ reana-client list --format run_number,name
RUN_NUMBER   NAME                   
1            reana-demo-root6-roofit
$ go run . list --format run_number,name
 NAME                     RUN_NUMBER
 reana-demo-root6-roofit  1    
```

Given that this PR is already pretty big, we can address this in another issue. It is probably a good idea to do so after discussing about using external libraries to manage tables.

cmd/list.go
```diff
@@ -289,3 +310,36 @@ func getOptionalIntField(value int64) string {
 	}
 	return fmt.Sprintf("%d", value)
 }
+
+func sortListData(data [][]any, header []string, sortColumn string) {
+	sortColumn = strings.ToLower(sortColumn)
+	if !slices.Contains(header, sortColumn) {
+		fmt.Println("Warning: column '" + sortColumn + "' does not exist")
+		return
+	}
```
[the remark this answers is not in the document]
BrunoRosendo: That's a great idea, it'll make the code flow much better. I think this should be implemented for #16. The code improvement of #41 could also be done there I think (otherwise the PR would be blocked for some time)

cmd/list.go
BrunoRosendo: Sounds good. For now, I'll just print the error and exit. This can also be improved in #16

utils/utils.go
```diff
@@ -126,6 +126,52 @@ func ParseListFilters(filter []string, filterNames []string) ([]string, string)
 	return statusFilters, searchFiltersString
 }
 
+func ParseFormatParameters(filters []string) map[string]any {
```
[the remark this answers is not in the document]
BrunoRosendo: Because I was using nil to know there was no filter. However, we can also use an empty string for that

utils/utils.go
BrunoRosendo: Created #44

mdonadoni (approved):
Thank you for your changes! We will handle the remaining little issues in the next PRs

PR #41, opened 2022-07-20
cli: add a way to control the output stream

### LINKED ISSUE #24: cli: add a way to control the output stream
Right now in mostly all the commands we simply print everything to the standard output using the `fmt` package ([example](https://github.com/reanahub/reana-client-go/blob/master/cmd/list.go#L112-L120)).
However to be able to test the output from commands we might need to control the stream and set it to a buffer before the command execution so that it can later be inspected inside the test. This was implemented in https://github.com/reanahub/reana-client-go/pull/10 and already works with `version` tests by setting the output stream to a buffer [here](https://github.com/reanahub/reana-client-go/blob/master/utils/utils.go#L23-L32).

One way to solve it would be to use [cobra Println command wrappers](https://pkg.go.dev/github.com/spf13/cobra#Command.Println) instead of `fmt`. In other words to use `cmd.Println()` instead of `fmt.Println()`, but the only problem with this approach is that cobra command wrappers uses [stderr by default](https://github.com/spf13/cobra/blob/v1.5.0/command.go#L1224)..

It might be confusing to some users if we print everything to `stderr`. Some projects solve it by setting the stream to `stdout` in `main.go` and pass it to the `rootCmd` and all the subcommands like [here](https://github.com/helm/helm/blob/c137bfb68b4f8b7a10687f043b0dbccb09574969/cmd/helm/helm.go#L66). And while testing do the same with the buffer, like [here](https://github.com/helm/helm/blob/49819b4ef782e80b0c7f78c30bd76b51ebb56dc8/cmd/helm/helm_test.go#L100).

We should investigate the alternative approaches, pick the best way to do it and implement it in all the commands.

BrunoRosendo:
closes #24

The strategy was:
- Change the Cobra's output to `os.Stdout`, since it uses `Stderr` by default. This only needed to be done at the root command
- In every command, change the `fmt` commands to `cmd` (e.g. `cmd.PrintLn()`)
- In `utils` or other functions without accces to `cmd`, pass the output through a parameter and use it

### DIFF
--- a/cmd/close.go
+++ b/cmd/close.go
@@ -52,7 +52,7 @@ func newCloseCmd() *cobra.Command {
 			if err := validation.ValidateWorkflow(workflow); err != nil {
 				return err
 			}
-			if err := close(token, workflow); err != nil {
+			if err := close(cmd, token, workflow); err != nil {
 				return err
 			}
 			return nil
@@ -66,7 +66,7 @@ func newCloseCmd() *cobra.Command {
 	return cmd
 }
 
-func close(token string, workflow string) error {
+func close(cmd *cobra.Command, token string, workflow string) error {
 	closeParams := operations.NewCloseInteractiveSessionParams()
 	closeParams.SetAccessToken(&token)
 	closeParams.SetWorkflowIDOrName(workflow)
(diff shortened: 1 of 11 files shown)


BrunoRosendo:
Converted to draft because I think it's better to implement this here, in the utils functions: https://github.com/reanahub/reana-client-go/pull/39#discussion_r926683191

BrunoRosendo:
Review after #47

mdonadoni (changes requested):
Works well! However the `open` and `close` commands are still using `fmt.Print*`. Could you please update the PR to also include the changes to those commands?

mdonadoni (approved):
Nice job!

PR #47, opened 2022-07-22
cli: improve error handling in the command functions

### LINKED ISSUE #46: cli: improve error handling in the command functions
Currently, every function detects its own errors, prints them to the standard output and exits the program. Instead, we should do the same thing that a lot of go libraries do: return the error to the calling function. We can do this in every function and use the cobra's `RunE` functionality, which will take care of writing the error to the user

BrunoRosendo:
Changes the logic we use to handle errors. Now, instead of simply printing an error to the console and exiting the program, the errors are returned to the calling function.
In the end, the error is handled by Cobra, using the RunE function option.

closes #46

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,7 +2,7 @@ package client
 
 import (
 	"crypto/tls"
-	"fmt"
+	"errors"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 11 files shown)


BrunoRosendo:
Do you think it's worth to return errors in [this function](https://github.com/reanahub/reana-client-go/blob/cd7a30db7691610c97f3200dfa4c102b1ed96db1/client/client.go#L24) as well? I don't think so because it's always a fatal error and it's a bit annoying to handle it in the commands, like [here](https://github.com/reanahub/reana-client-go/blob/cd7a30db7691610c97f3200dfa4c102b1ed96db1/cmd/du.go#L99)

BrunoRosendo:
> Do you think it's worth to return errors in [this function](https://github.com/reanahub/reana-client-go/blob/cd7a30db7691610c97f3200dfa4c102b1ed96db1/client/client.go#L24) as well? I didn't think so because it's always a fatal error and it's a bit annoying to handle it in the commands, like [here](https://github.com/reanahub/reana-client-go/blob/cd7a30db7691610c97f3200dfa4c102b1ed96db1/cmd/du.go#L99)

On a second thought, I think it may be worth it, since otherwise we'd have to always pass the output stream to the function. But I'd like some feedback on it

mdonadoni (approved):
LGTM!

PR #50, opened 2022-07-26
cli: fix list filter by replacing stringArray with stringSlice

### LINKED ISSUE #31: cli: implement `logs` command
Create a new `logs` command that replicates all the functionality we have in `reana-client`.
It should also include all the flags and return exactly the same output as `reana-client logs`.

Please note that `status` command implementeed in `reana-client` uses some coloring which could be added later as part of https://github.com/reanahub/reana-client-go/issues/2 issue.

BrunoRosendo:
closes #31


mdonadoni (approved):
LGTM!

Issue #2, opened 2022-06-21
cli: introduce a logging library


audrium:
Introduce some nice info/debug/error logging library with coloring like we have in [python client](https://github.com/reanahub/reana-client/blob/master/reana_client/printer.py#L21) and implement a logic for a [loglevel](https://github.com/reanahub/reana-client-go/pull/1/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR33) flag

_Originally posted by @tiborsimko in https://github.com/reanahub/reana-client-go/pull/1#discussion_r901818913_

Libraries to consider:
- PTerm: https://pterm.sh/

audrium:
Nice findings! I've checked [Logrus](https://github.com/sirupsen/logrus) and it seems that it's exactly what we need to log all the actions we perform and to implement a [loglevel](https://github.com/reanahub/reana-client-go/pull/1/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR33) flag. And as you said [Go-Pretty](https://github.com/jedib0t/go-pretty) could be used to implement something similar to `display_message()` utility for coloring the output!

PR #54, opened 2022-07-29
cli: implement loglevel flag

audrium:
closes https://github.com/reanahub/reana-client-go/issues/2

Implements a global `loglevel` flag by using a [Logrus](https://github.com/sirupsen/logrus) library.

To test:
```
- go run . du -w workflow -l DEBUG # should display the output from go-swagger and print all the flags used in the cmd
- go run . open -w workflow -l INFO # should display only an info level message
- go run . du -w workflow -l TEST # should return an error
```

As part of this PR changed `ValidateArgChoice()` to more generic `ValidateChoice()` to allow using it also for flags.


utils/utils.go
```diff
@@ -203,3 +205,10 @@ func GetWorkflowNameAndRunNumber(workflowName string) (string, string) {
 func FormatSessionURI(serverURL string, path string, token string) string {
 	return serverURL + path + "?token=" + token
 }
+
+func LogCmdFlags(cmd *cobra.Command) {
```
mdonadoni, line 209: What do you think about calling this function in the `runE` function of the root command, so that it is automatically done for each command?

cmd/root.go
```diff
@@ -20,11 +22,14 @@ func NewRootCmd() *cobra.Command {
...
 		SilenceUsage: true,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			return runE(cmd)
+		},
 	}
 
 	cmd.SetOut(os.Stdout)
 
-	cmd.PersistentFlags().BoolP("loglevel", "l", false, "Sets log level [DEBUG|INFO|WARNING]")
+	cmd.PersistentFlags().StringP("loglevel", "l", "WARNING", "Sets log level [DEBUG|INFO|WARNING]")
```
mdonadoni, line 32: What do you think about using a struct to save this flag, just like we did for the other commands in https://github.com/reanahub/reana-client-go/pull/56?

cmd/root.go
```diff
@@ -38,3 +43,33 @@ func NewRootCmd() *cobra.Command {
...
+func runE(cmd *cobra.Command) error {
+	logLevelFlag, _ := cmd.Flags().GetString("loglevel")
+
+	if err := addLogger(logLevelFlag); err != nil {
+		return err
+	}
+	return nil
+}
+
+func addLogger(logLevelFlag string) error {
```
mdonadoni, line 56:
```suggestion
func setupLogger(logLevelFlag string) error {
```
WDYT?

cmd/root.go
```diff
@@ -38,3 +43,33 @@ func NewRootCmd() *cobra.Command {
...
 	return cmd
 }
+
+func runE(cmd *cobra.Command) error {
+	logLevelFlag, _ := cmd.Flags().GetString("loglevel")
+
+	if err := addLogger(logLevelFlag); err != nil {
+		return err
+	}
+	return nil
```
mdonadoni, line 53:
```suggestion
	return addLogger(logLevelFlag); 
```
WDYT?

cmd/root.go
```diff
@@ -38,3 +43,33 @@ func NewRootCmd() *cobra.Command {
...
+		"loglevel",
+	); err != nil {
+		return err
+	}
+	level, err := log.ParseLevel(logLevelFlag)
+	if err != nil {
+		return err
+	}
+	log.SetLevel(level)
+	log.SetOutput(os.Stdout)
```
mdonadoni, line 69: What do you think about writing to `os.Stderr`, just like the Python client does?

utils/utils.go
audrium: Nice suggestion, thanks

cmd/root.go
audrium: I left this as it is, since here we could add different things like viper configuration, etc and this syntax could be useful. Please see https://github.com/reanahub/reana-client-go/pull/60/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR49-R65

cmd/root.go
audrium: Nice catch

mdonadoni (approved):
Looks good, well done!

PR #56, opened 2022-07-29
cli: use options struct for flags and configuration

### LINKED ISSUE #49: cli: handle flag parsing errors
Command flags are defined in the function responsible for creating a new command (e.g. `newListCmd`) and are then parsed in the function that implements the logic of the command (e.g. `list`)

https://github.com/reanahub/reana-client-go/blob/06576d7027467cef61c6c7891900d00ee15bf8f9/cmd/list.go#L83

https://github.com/reanahub/reana-client-go/blob/06576d7027467cef61c6c7891900d00ee15bf8f9/cmd/list.go#L107

The `Get*` methods are used to get the flags' values, however there can be errors when the method used does not match the type of the flag (e.g. using `GetStringArray` for a `StringSlice` flag). For this reason, these methods might return an error.

Currently we do not check these errors, but this is not a good idea as flags might be unexpectedly ignored (see #48) .

Two possible options:
1. Check the errors returned by `Get*`. However, this would make the code much more verbose.
2. Define the flags using the `*Var` variants (e.g. `StringVar`). This is what GitHub CLI and Helm do. For example, this is the struct used to store the flags' values of the `helm repo add` command ([source](https://github.com/helm/helm/blob/937798868518b7c5dbcbefac5bfe0b2e045e3bae/cmd/helm/repo_add.go#L46-L66)):

    ```go
    type repoAddOptions struct {
    	name                 string
    	url                  string
    	username             string
    	password             string
    	...
    }
    ```

    These are the definitions of the flags ([source](https://github.com/helm/helm/blob/937798868518b7c5dbcbefac5bfe0b2e045e3bae/cmd/helm/repo_add.go#L69-L97)):
    ```go
    o := &repoAddOptions{}
    
    cmd := &cobra.Command{
    	...
    }
    
    f := cmd.Flags()
    f.StringVar(&o.username, "username", "", "chart repository username")
    f.StringVar(&o.password, "password", "", "chart repository password")
    ...
    ```
    The flags' values can then be read from the defined variable, instead of calling `Get*`.

BrunoRosendo:
closes #49

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -37,72 +37,77 @@ interactive session. Overrides the default image
 for the selected type.
 `
 
+type openOptions struct {
+	token                  string
+	serverURL              string
+	workflow               string
+	image                  string
+	interactiveSessionType string
+}
+
 func newOpenCmd() *cobra.Command {
+	o := &openOptions{}
+
 	cmd := &cobra.Command{
 		Use:   "open",
 		Short: "Open an interactive session inside the workspace.",
 		Long:  openDesc,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			token, _ := cmd.Flags().GetString("access-token")
-			if token == "" {
-				token = os.Getenv("REANA_ACCESS_TOKEN")
+			if o.token == "" {
+				o.token = os.Getenv("REANA_ACCESS_TOKEN")
 			}
-			serverURL := os.Getenv("REANA_SERVER_URL")
-			workflow, _ := cmd.Flags().GetString("workflow")
-			if workflow == "" {
-				workflow = os.Getenv("REANA_WORKON")
+			o.serverURL = os.Getenv("REANA_SERVER_URL")
+			if o.workflow == "" {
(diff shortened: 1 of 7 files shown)


cmd/open.go
```diff
@@ -37,72 +37,77 @@ interactive session. Overrides the default image
...
 	}
 
-	cmd.Flags().StringP("access-token", "t", "", "Access token of the current user.")
-	cmd.Flags().
-		StringP("workflow", "w", "", "Name or UUID of the workflow. Overrides value of REANA_WORKON environment variable.")
-	cmd.Flags().StringP("image", "i", "", openImageFlagDesc)
+	f := cmd.Flags()
+	f.StringVarP(&o.token, "access-token", "t", "", "Access token of the current user.")
+	f.StringVarP(
+		&o.token,
```
mdonadoni, line 94:
```suggestion
		&o.workflow,
```

mdonadoni (changes requested):

mdonadoni (approved):
LGTM!

PR #57, opened 2022-08-01
cli: centralize logic that handles env var configuration

### LINKED ISSUE #45: cli: centralize logic that handles env variables configuration
In order to reduce repeated code, we should discuss and implement a way to centralize the logic involving configuration using environment variables. For example, take a look at [this comment](https://github.com/reanahub/reana-client-go/pull/42#discussion_r927389682)

BrunoRosendo:
closes #45

### DIFF
--- a/config/env.go
+++ b/config/env.go
@@ -1,0 +1,35 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package config
+
+import "os"
+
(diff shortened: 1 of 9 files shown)


config/env.go
```diff
@@ -0,0 +1,35 @@
...
+
+type envConfig struct {
+	ServerURL   string
+	AccessToken string
+	Workflow    string
+}
+
+func newEnvConfig() *envConfig {
+	config := &envConfig{
+		ServerURL:   envOrDefault("REANA_SERVER_URL", ""),
```
BrunoRosendo, line 23: This is the same as `os.GetEnv()` but I think this is more clear and we can use the same logic for other default values in the future

PR #59, opened 2022-08-01
cli: add display message utility

audrium:
closes https://github.com/reanahub/reana-client-go/issues/2

Built on top of [cli: implement loglevel flag](https://github.com/reanahub/reana-client-go/pull/54)

This PR introduces an utility similar to [display_message() in Python client](https://github.com/reanahub/reana-client/blob/master/reana_client/printer.py#L21) to colorize the output in `reana-client-go`

### DIFF
--- a/utils/display.go
+++ b/utils/display.go
@@ -14,7 +14,25 @@ import (
 	"io"
 
 	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
 )
+
+type MessageType int
+
+const (
+	Success MessageType = iota
+	Warning
+	Error
+	Info
+)
+
+func (m MessageType) String() string {
+	return []string{"SUCCESS", "WARNING", "ERROR", "INFO"}[m]
+}
+
+func (m MessageType) Color() text.Color {
+	return []text.Color{text.FgGreen, text.FgYellow, text.FgRed, text.FgCyan}[m]
+}
 
 func DisplayTable(header []string, rows [][]any, out io.Writer) {
 	// Convert to table.Row type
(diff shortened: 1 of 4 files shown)


utils/display.go
```diff
@@ -51,3 +54,35 @@ func DisplayJsonOutput(output any, out io.Writer) error {
...
+	}
+	fmt.Fprintf(out, "%s%s\n", prefixTpl, message)
+}
+
+func colorizeMessage(msg string, msgType string) string {
+	colors := map[string]text.Color{
+		"success": text.FgGreen,
+		"warning": text.FgYellow,
+		"error":   text.FgRed,
+		"info":    text.FgCyan,
```
mdonadoni, line 85: (minor) What about using an enum for the message type, both here and in `DisplayMessage`? Go doesn't really have enums, but something similar can be done like this: https://github.com/jedib0t/go-pretty/blob/eb654fb730581a1a87290d9b1087867232a7b314/text/color.go#L33-L48

audrium: Tried to quickly implement something, but it looked a bit messy, since we have to define an enum and still a map to match the values

mdonadoni: What I wanted to achieve was to avoid having to use strings to call the function, that is having to pass `"success"` or one of the other options, as it is quite error prone (e.g. typos that go unnoticed). Maybe we can at least define them as constants?

Something like this:
```go
type MessageType string

const (
    Success MessageType = "success"
    ...
)
```

Inside the functions we can just convert them back to strings, so that we do not have to change the functions' code. What do you think? Are there any other cons I am missing?

audrium: Thanks for the feedback, I think I managed to find a nice solution, please have a look

mdonadoni: I think this solution is great!

mdonadoni (approved):
LGTM! Well done!

PR #60, opened 2022-08-02
cli: handle env var configuration and validation

### LINKED ISSUE #45: cli: centralize logic that handles env variables configuration

audrium:
closes #45
 
 Built on top of https://github.com/reanahub/reana-client-go/pull/59
 
 Introduces viper library to bind and set environment variables to the flags. Also added a way to validate flags in `PersistentPreRunE`

To test make sure that setting a flag overrides envar value:
```
$ REANA_WORKON="value1" go run . du -w value2 # value2 should be used
```


cmd/root.go
```diff
@@ -57,10 +59,68 @@ func (o *rootOptions) run(cmd *cobra.Command) error {
...
 	return nil
 }
 
+func validateFlags(cmd *cobra.Command) error {
+	token := cmd.Flags().Lookup("access-token")
+	serverURL := viper.GetString("server-url")
+	workflow := cmd.Flags().Lookup("workflow")
+
+	if token != nil {
+		tokenValue := fmt.Sprintf("%v", token.Value)
```
mdonadoni, line 80:
```suggestion
		tokenValue := token.Value.String()
```
WDYT?

cmd/root.go
```diff
@@ -57,10 +59,68 @@ func (o *rootOptions) run(cmd *cobra.Command) error {
...
+		if err := validation.ValidateAccessToken(tokenValue); err != nil {
+			return err
+		}
+		if err := validation.ValidateServerURL(serverURL); err != nil {
+			return err
+		}
+	}
+
+	if workflow != nil {
+		workflowValue := fmt.Sprintf("%v", workflow.Value)
```
mdonadoni, line 90:
```suggestion
		workflowValue := workflow.Value.String()
```

cmd/list.go
```diff
@@ -82,21 +81,8 @@ func newListCmd() *cobra.Command {
...
-				return err
-			}
-			if err := validation.ValidateServerURL(o.serverURL); err != nil {
-				return err
-			}
-			if err := o.run(cmd); err != nil {
-				return err
-			}
-			return nil
+			o.serverURL = viper.GetString("server-url")
```
mdonadoni, line 84: On an unrelated note, it might be a good idea to also have a command flag for the server URL in the future. Do you know if there is a reason why we do not do it in the Python client?

mdonadoni (commented):
Looks good! We don't use many of viper's features, but they might become useful in the future

cmd/list.go
audrium: Indeed I had the same question why there is a flag for a token and not for a server url. @tiborsimko, do you remember the reason?

cmd/root.go
```diff
@@ -60,11 +61,70 @@ func (o *rootOptions) run(cmd *cobra.Command) error {
...
+			return err
+		}
+	}
+
+	if workflow != nil {
+		workflowValue := workflow.Value.String()
+		if err := validation.ValidateWorkflow(workflowValue); err != nil {
+			return err
+		}
+	}
```
mdonadoni, line 97: I have an issue with the `list` command:
```console
$ ./reana-client-go list 
Error: workflow name must be provided either with `--workflow` option or with REANA_WORKON environment variable
```
Can you reproduce this?

audrium: Nice find! Indeed `--workflow` flag for `list` command is different compared to the other commands. It's optional and should not be validated in this way

audrium: Fixed it by manually setting the `--workflow` flag for `list` command to be `Changed` (as if the user would set the value) to skip the validation and setting `REANA_WORKON` environment variable value. A bit hacky solution, but could not come up with anything cleaner. Alternative would be to handle `--workflow` validation and setting of `REANA_WORKON` value in each of the commands which is not great either

mdonadoni: I haven't investigated much, but it seems there is a way to [mark a flag as required](https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired). Maybe using this it's possible to set in each command whether the flag is required or not, but only do the actual validation in a single place inside `validateFlags`?

audrium: After playing a lot with cobra and pflag api and trying various options similar to what you mentioned I came up with one solution which might be the cleanest. It's based on the flag usage information, since for required workflow flags it's all the same as [here](https://github.com/reanahub/reana-client/blob/e1971cac6bb039053037c1be9e2c0958eb7e0558/reana_client/cli/utils.py#L116). Please have a look

mdonadoni: For sure it works, but it still is an hacky solution. Have you also looked at annotations (see [`SetAnnotation`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.SetAnnotation) and [`Flag.Annotations`](https://pkg.go.dev/github.com/spf13/pflag#Flag))? Maybe we can use them to annotate the `workflow` flag in `list`, and then we can check that when doing binding/validation. The annotation can be something like `do_not_bind_to_env` or similar. This would replace the check done on the usage string. What do you think?

audrium: Tried annotations with the older approach and it was very verbose to use it, but now it fits quite nicely. I think it's better than relying on the usage string indeed

mdonadoni (approved):
LGTM!

PR #61, opened 2022-08-02
cli: use dataframe library to manipulate tabular data

### LINKED ISSUE #43: cli: Replace table logic with an external library
Table data is and will be strongly present in the client. Currently, this logic is being handled manually with slices. It would be good to use an external library (maybe dataframes or something like [tablib](https://tablib.readthedocs.io/en/stable/))

Replace any existing code that can be improved by this

### LINKED ISSUE #44: cli: ensure `format` flag respects the order passed in the command
Just like discussed in #39, the `format` flag currently does not respect the order of headers that is passed in the flag. For example:

**Expected:**
```
$ reana-client list --format run_number,name
RUN_NUMBER   NAME                   
1            reana-demo-root6-roofit
```

**Currently:**
```
$ go run . list --format run_number,name
 NAME                     RUN_NUMBER
 reana-demo-root6-roofit  1    
 ```

### PR DESCRIPTION
**Pull request by BrunoRosendo**
closes #43 
closes #44

### CODE REVIEW

File: cmd/list.go [hunk 1]
```diff
@@ -248,12 +245,12 @@ func displayListPayload(
 				if humanReadable {
 					value = workflow.Size.HumanReadable
 				} else {
-					value = workflow.Size.Raw
+					value = int(workflow.Size.Raw)
```
BrunoRosendo, line 248: This cast was needed because the library doesn't support `int64`. Any ideas of improvements would be welcome

audrium (approved):
Very nice work!
P.S. rebase is needed

PR #64, opened 2022-08-04
cli: standardize the way filters are being handled in different commands

### LINKED ISSUE #53: cli: standardize the way filters are being handled in different commands
As discussed in [this comment](https://github.com/reanahub/reana-client-go/pull/51#discussion_r930838279), the filters are currently being handled by different ways in different commands (e.g., list uses `utils.ParseFilterParameters` while logs uses `parseLogsFilters`). This happens because they have different filters that have slight differences on their validation and work in the same way as the Python client.

However, we could have a standard and general way of handling filters. For example, we could define a struct `Filters` with functions to validate them given a list of filters, to get them (e.g `get("key")` or `getStatus()`), among others that might prove useful.

BrunoRosendo:
closes #53


cmd/du.go
```diff
@@ -103,7 +103,12 @@ func newDuCmd() *cobra.Command {
 
 func (o *duOptions) run(cmd *cobra.Command) error {
 	filterNames := []string{"size", "name"}
-	_, searchFilter, err := utils.ParseFilterParameters(o.filter, filterNames)
+	filters := utils.NewFilters(nil, filterNames)
```
audrium, line 106: Minor: maybe `NewFilters()` could also accept the filter values (`o.filter`) and call `AddFilters()` internally to initialize?
Or do you think there is a reason to keep the two steps separated?

I mean it could be used like this:
```go
filters, err := utils.NewFilters(nil, filterNames, o.filter)
if err != nil {
    return err
}
```

audrium (commented):
Very nice work, it works as expected!
Left couple of minor suggestions, but it looks good to me!

cmd/du.go
BrunoRosendo: I agree, it becomes less verbose this way

audrium (approved):
LGMT

PR #65, opened 2022-08-05
cli: implement status command

### LINKED ISSUE #30: cli: implement `status` command
Create a new `status` command that replicates all the functionality we have in `reana-client`.
It should also include all the flags and return exactly the same output as `reana-client status`.

Example:
```
$ reana-client status -w workflow
NAME       RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS     PROGRESS
workflow   1            2022-07-06T07:40:23   2022-07-06T07:40:34   2022-07-06T07:40:50   finished   2/2

```

BrunoRosendo:
closes #30

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -1,0 +1,261 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/utils"
+	"strings"
+
+	"golang.org/x/exp/slices"
+
+	"github.com/go-gota/gota/dataframe"
+	"github.com/go-gota/gota/series"
+
+	"github.com/spf13/cobra"
+)
+
+const statusDesc = `
+Get status of a workflow.
+
+The ` + "``status``" + ` command allow to retrieve status of a workflow. The status
+can be created, queued, running, failed, etc. You can increase verbosity or
+filter retrieved information by passing appropriate command-line options.
+
+Examples:
+
+  $ reana-client status -w myanalysis.42
+
+  $ reana-client status -w myanalysis.42 -v --json
+`
+
+const statusFormatFlagDesc = `Format output by displaying only certain columns.
+E.g. --format name,status.`
+
+type statusOptions struct {
+	token           string
+	workflow        string
+	formatFilters   []string
+	jsonOutput      bool
+	verbose         bool
+	includeDuration bool
+}
+
+// newStatusCmd creates a command to get status of a workflow.
+func newStatusCmd() *cobra.Command {
+	o := &statusOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "status",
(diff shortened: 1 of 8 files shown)


cmd/status.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	payload := statusResp.Payload
+
+	header := buildStatusHeader(
+		o.verbose,
+		o.includeDuration,
+		payload.Progress,
+		payload.Status,
+	)
+	parsedFormatFilters := utils.ParseFormatParameters(o.formatFilters, false)
```
BrunoRosendo, line 132: While doing this, I noticed the Python client doesn't respect order in the format filter. Should we open an issue for that?

utils/filters.go
```diff
@@ -193,12 +193,12 @@ type FormatFilter struct {
...
 // ParseFormatParameters parses a list of formatOptions to a slice of FormatFilter.
 // If the format option has a filter, that will be the value in the struct and the filterRows boolean will be true.
-func ParseFormatParameters(formatOptions []string) []FormatFilter {
+func ParseFormatParameters(formatOptions []string, filterRows bool) []FormatFilter {
 	var parsedFilters []FormatFilter
 	for _, filter := range formatOptions {
 		filterNameAndValue := strings.SplitN(filter, "=", 2)
 		formatFilter := FormatFilter{column: filterNameAndValue[0], filterRows: false}
-		if len(filterNameAndValue) >= 2 {
+		if filterRows && len(filterNameAndValue) >= 2 {
```
mdonadoni, line 201: Should we return an error if `!filterRows` and `len(filterNameAndValue) >= 2`?

cmd/status.go
mdonadoni: Yes please!

cmd/status.go
```diff
@@ -0,0 +1,261 @@
...
+Examples:
+
+  $ reana-client status -w myanalysis.42
+
+  $ reana-client status -w myanalysis.42 -v --json
+`
+
+const statusFormatFlagDesc = `Format output by displaying only certain columns.
+E.g. --format name,status.
+`
```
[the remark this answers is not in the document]
BrunoRosendo: I agree, I'll change this in other commands too

utils/filters.go
BrunoRosendo: Hmm I'm not sure. For example, in the status command, the format does not filter. That said, if the user inserts `--format name=root`, should we return an error because of that? We could just ignore everything after `=`, that's the current behavior in the Python client. Do you think this should be changed?

Otherwise, do we return an error because it includes an `=` or because the string `name=root` isn't a valid filter?

mdonadoni: It just feels odd that the user can specify `--format name=root` and we simply ignore `root`, but I guess we can leave it like this for compatibility with r-client 😄

cmd/status.go
```diff
@@ -0,0 +1,261 @@
...
+			if err != nil {
+				return err
+			}
+		}
+
+		colSeries.Append(value)
+		df = df.CBind(dataframe.New(colSeries))
+	}
+
+	df = utils.FormatDataFrame(df, filters)
```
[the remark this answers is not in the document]
BrunoRosendo: I opened #69, should we open one for r-client too? :sweat_smile:

mdonadoni (approved):

PR #66, opened 2022-08-05
cli: implement ls command

### LINKED ISSUE #20: cli: implement `ls` command
Create a new `ls` command that replicates all the functionality we have in `reana-client`.
It should also include all the flags like `url`, `human-readable` and pagination flags and return exactly the same output as `reana-client ls`.

Example:
```
$ reana-client ls -w workflow
NAME                SIZE     LAST-MODIFIED
results/data.root   154458   2022-07-06T07:40:36
results/plot.png    15450    2022-07-06T07:40:43
code/gendata.C      1937     2022-07-06T07:40:23
code/fitdata.C      1648     2022-07-06T07:40:23
```

BrunoRosendo:
closes #20

This openapi changes also contain the ones in #65 so merge it afterwards

### DIFF
--- a/cmd/ls.go
+++ b/cmd/ls.go
@@ -1,0 +1,233 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/utils"
+
+	"github.com/go-gota/gota/dataframe"
+	"github.com/go-gota/gota/series"
+	"github.com/spf13/viper"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/spf13/cobra"
+)
+
+const lsDesc = `
+List workspace files.
+
+The ` + "``ls``" + ` command lists workspace files of a workflow specified by the
+environment variable REANA_WORKON or provided as a command-line flag
+` + "``--workflow`` or ``-w``." + ` The SOURCE argument is optional and specifies a
+pattern matching files and directories.
+
+Examples:
+
+  $ reana-client ls --workflow myanalysis.42
+
+  $ reana-client ls --workflow myanalysis.42 --human-readable
+
+  $ reana-client ls --workflow myanalysis.42 'data/*root*'
+
+  $ reana-client ls --workflow myanalysis.42 --filter name=hello
+`
+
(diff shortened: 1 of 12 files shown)


cmd/ls.go
```diff
@@ -0,0 +1,232 @@
...
+func buildLsSeries(col string, humanReadable bool) series.Series {
+	if col == "size" && !humanReadable {
+		return series.New([]int{}, series.Int, col)
+	}
+	return series.New([]string{}, series.String, col)
+}
+
+func displayLsURLs(cmd *cobra.Command, p *operations.GetFilesOKBody, serverURL string, workflow string) {
+	for _, file := range p.Items {
+		fileURL := fmt.Sprintf("%s/api/workflows/%s/workspace/%s", serverURL, workflow, file.Name)
```
BrunoRosendo, line 229: Python client retrieves the URL from the swagger api but I didn't find a way to do it here, so I wrote it like this

cmd/ls.go
```diff
@@ -0,0 +1,237 @@
...
+
+  $ reana-client ls --workflow myanalysis.42 --filter name=hello
+`
+
+const lsFormatFlagDesc = `
+Format output according to column titles or column
+values. Use <column_name>=<column_value> format.
+E.g.display files named data.txt
+--format name=data.txt
+`
```
mdonadoni, line 50: I think we should remove the newline at the beginning and at the end of this string, so that they are not present in the help message. What do you think?

cmd/ls.go
```diff
@@ -0,0 +1,237 @@
...
+
+const lsFilterFlagDesc = `
+Filter results to show only files that match certain filtering criteria such as
+file name, size or modification date.
+Use --filter <column_name>=<column_value> pairs. Available
+filters are 'name', 'size' and 'last-modified'.
+`
```
mdonadoni, line 57: Same as above

cmd/ls.go
mdonadoni: I don't think there is a way to do this right now. I also think that this is not needed in many places, so perhaps we can keep it like this and we can open an issue just for future reference?

cmd/ls.go
```diff
@@ -0,0 +1,237 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "ls",
+		Short: "List workspace files.",
+		Long:  lsDesc,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			o.serverURL = viper.GetString("server-url")
+			if len(args) > 0 {
+				o.fileName = args[0]
+			}
```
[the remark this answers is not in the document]
BrunoRosendo: That's very useful, thanks!

mdonadoni (approved):
Nice job!

Issue #62, opened 2022-08-03
cli: add logs and colours to the commands that are missing them


BrunoRosendo:
Currently, not every command has coloured messages (from the `DisplayMessage` utils function) like the Python client has. There are also some logs missing.
This happened because those tasks were made in parallel and this was not taken into consideration in the rebases. The task should basically involve comparing the output to the Python client and adding colours or logs where it's needed.

PR #71, opened 2022-08-08
cli: add missing logs and messages colors

BrunoRosendo:
Adds logs and colors to the messages that were missing them, according mostly to the Python client.

closes https://github.com/reanahub/reana-client-go/issues/62

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -14,6 +14,8 @@ import (
 	"reanahub/reana-client-go/utils"
 	"reanahub/reana-client-go/validation"
 
+	"github.com/jedib0t/go-pretty/v6/text"
+
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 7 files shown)


cmd/open.go
```diff
@@ -107,7 +109,8 @@ func (o *openOptions) run(cmd *cobra.Command) error {
 		false,
 		cmd.OutOrStdout(),
 	)
-	cmd.Println(utils.FormatSessionURI(o.serverURL, openResp.Payload.Path, o.token))
+	sessionURI := utils.FormatSessionURI(o.serverURL, openResp.Payload.Path, o.token)
+	utils.PrintColorable(sessionURI, cmd.OutOrStdout(), text.FgGreen)
```
mdonadoni, line 113: The newline is missing at the end of this line

cmd/ping.go
```diff
@@ -71,5 +75,8 @@ func (o *pingOptions) run(cmd *cobra.Command) error {
 		fmt.Sprintf("Status: %s ", "Connected")
 
 	cmd.Println(response)
+	res, _ := p.MarshalBinary()
+	log.Debug("Server response:\n", string(res))
+
```
mdonadoni, line 80: I think we do not need this, given that the response of each request is already printed here:
https://github.com/reanahub/reana-client-go/blob/24440c20747488bc06118872d3c56b0b42544110/client/client.go#L49-L50

mdonadoni (approved):
Works great!

PR #72, opened 2022-08-09
cli: unit tests for validation/utils.go and utils/utils.go

### LINKED ISSUE #70: cli: create tests for existing functionalities and commands

- [x] Utils functions
- [x] Validation functions
- [x] `close` command
- [x] `du` command
- [x] `info` command
- [ ] `list` command
- [x] `logs` command
- [x] `open` command
- [ ] `status` command
- [x] `ls` command

BrunoRosendo:
closes #70

### DIFF
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -1,0 +1,248 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package utils
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/spf13/viper"
+
+	"golang.org/x/exp/slices"
+)
+
+func TestGetRunStatuses(t *testing.T) {
+	tests := []struct {
+		includeDeleted bool
+		numStatuses    int
+	}{
+		{includeDeleted: false, numStatuses: 7},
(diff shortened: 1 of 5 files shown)


BrunoRosendo:
Currently, LogCmdFlags and BindViperToCmdFlag are not being tested. Do you think it's worth it to introduce a mocking library to test this type of functions? We could test if they right methods are being called with given parameters

I was checking some libraries and found 2 good options: [GoMock](https://github.com/golang/mock) and [testify](https://github.com/stretchr/testify)

utils/utils_test.go
```diff
@@ -0,0 +1,248 @@
...
+			if !slices.Contains(runStatuses, "deleted") {
+				t.Errorf("Expected runStatuses to contain deleted")
+			}
+		} else if slices.Contains(runStatuses, "deleted") {
+			t.Errorf("Expected runStatuses not to contain deleted")
+		}
+	}
+}
+
+func TestHasAnyPrefix(t *testing.T) {
```
tiborsimko, line 47: It's great to see many tests being added to the Go client. Can you please also add the code coverage information to in the output? IOW, just edit `Makefile` and use `go test -cover ./...` there, with the option `-cover`.  

(Note also that I think we can remove `-v` in that command, because it will make a nicer formatting with coverage without it, and if some test fails, then the failed test will be printed out in full anyway.)

BrunoRosendo: That sounds very good, thanks

mdonadoni (approved):

PR #73, opened 2022-08-09
tests: function to test commands and unit tests for root.go

### LINKED ISSUE #70: cli: create tests for existing functionalities and commands

BrunoRosendo:
closes #70 

Review after #72

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1,0 +1,65 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"reanahub/reana-client-go/utils"
+	"strings"
+	"testing"
+
+	"github.com/spf13/viper"
+)
+
+func testCmdRun(
+	t *testing.T,
+	cmd, serverPath, serverResponse string,
+	wantError bool, expectedMsgs []string,
+	args ...string,
+) {
+	server := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if accessToken := r.URL.Query().Get("access_token"); accessToken != "1234" {
+			t.Errorf("Expected access token '1234', got '%v'", accessToken)
+		}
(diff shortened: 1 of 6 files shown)


cmd/root_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/spf13/viper"
+)
+
+func testCmdRun(
+	t *testing.T,
+	cmd, serverPath, serverResponse string,
+	wantError bool, expectedMsgs []string,
+	args ...string,
+) {
```
mdonadoni, line 30: This function has quite a lot of parameters and it currently only support one endpoint, but we can improve this later when needed!

cmd/root_test.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	for _, test := range tests {
+		cmd := NewRootCmd()
+		f := cmd.Flags()
+		if test.hasToken {
+			f.String("access-token", test.token, "")
+		}
+		if test.hasServerURL {
+			viper.Set("server-url", test.serverURL)
```
mdonadoni, line 129: I think we need to call `viper.Reset()` at the end of each subtest, don't we?

mdonadoni:
There is also a small typo in the commit title `cli: create unit tests tests for root.go` (the word `tests` is repeated two times). Should we also use `tests: ...` instead of `cli: ...`?

cmd/root_test.go
```diff
@@ -63,3 +67,146 @@ func testCmdRun(
...
+		viperProp string
+		value     string
+	}{
+		{env: "REANA_SERVER_URL", viperProp: "server-url", value: "https://localhost:8080"},
+		{env: "REANA_ACCESS_TOKEN", viperProp: "access-token", value: "1234"},
+		{env: "REANA_WORKON", viperProp: "workflow", value: "workflow"},
+	}
+
+	for _, test := range tests {
+		err := os.Setenv(test.env, test.value)
```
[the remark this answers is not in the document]
BrunoRosendo: You're right. I'm doing the `t.Run(...)` changed in another PR, so I'm doing this there

mdonadoni (approved):
Nice job!

PR #74, opened 2022-08-10
tests: unit tests for utils/filters.go and name subtests

### LINKED ISSUE #70: cli: create tests for existing functionalities and commands

BrunoRosendo:
closes #70

### DIFF
--- a/utils/filters_test.go
+++ b/utils/filters_test.go
@@ -16,34 +16,34 @@ import (
 )
 
 func TestNewFilters(t *testing.T) {
-	tests := []struct {
+	tests := map[string]struct {
 		singleFilterKeys []string
 		multiFilterKeys  []string
 		inputFilters     []string
 		testFilter       bool
 		wantError        bool
 	}{
-		{
+		"empty filters": {
 			singleFilterKeys: []string{},
 			multiFilterKeys:  []string{},
 		},
-		{
+		"no filter values": {
 			singleFilterKeys: []string{"single"},
 			multiFilterKeys:  []string{"multi"},
 		},
-		{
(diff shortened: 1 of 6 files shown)


utils/filters_test.go
```diff
@@ -0,0 +1,561 @@
...
+				}
+			}
+		})
+	}
+
+	filters, _ := NewFilters([]string{"single"}, []string{}, []string{})
+	err := filters.AddFilter("invalid_filter")
+	if err == nil {
+		t.Errorf("Expected error for AddFilter(%s), got nil", "invalid_filter")
+	}
```
mdonadoni, line 203: What do you think about using `t.Run(...)` also for this test?

utils/filters_test.go
```diff
@@ -0,0 +1,561 @@
...
+					if filter != "value" {
+						t.Errorf("Expected filter value to be 'value', got '%s'", filter)
+					}
+				}
+			}
+		})
+	}
+}
+
+func TestFilters_AddFilters(t *testing.T) {
```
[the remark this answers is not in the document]
BrunoRosendo: the IDE suggested this for some reason, but I also prefer without the `_`

utils/filters_test.go
BrunoRosendo: Nice catch

mdonadoni (approved):
Looks good!

PR #78, opened 2022-08-12
tests: unit tests for open and du commands

### LINKED ISSUE #70: cli: create tests for existing functionalities and commands

BrunoRosendo:
closes #70

### DIFF
--- a/cmd/open_test.go
+++ b/cmd/open_test.go
@@ -1,0 +1,77 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"net/http"
+	"reanahub/reana-client-go/utils"
+	"strings"
+	"testing"
+)
+
+var openPathTemplate = "/api/workflows/%s/open/%s"
+
+func TestOpen(t *testing.T) {
+	successResponse := `{"path": "/test/jupyter"}`
+	workflowName := "my_workflow"
+	tests := map[string]TestCmdParams{
+		"success default": {
+			serverPath: fmt.Sprintf(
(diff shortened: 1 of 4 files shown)


cmd/open_test.go
```diff
@@ -23,7 +23,11 @@ func TestOpen(t *testing.T) {
 	workflowName := "my_workflow"
 	tests := map[string]TestCmdParams{
 		"success default": {
-			serverPath:     fmt.Sprintf(openPathTemplate, workflowName, utils.InteractiveSessionTypes[0]),
+			serverPath: fmt.Sprintf(
+				openPathTemplate,
+				workflowName,
+				utils.InteractiveSessionTypes[0],
+			),
```
mdonadoni, line 30: Could you please move this change to the first commit?

mdonadoni (commented):
Looks good! After seeing a few tests for different commands, I think that it was a very good idea to implement and use the `testCmdRun` function!

In any case, I have seen that sometimes the commit messages are pretty long. Even though it is not a hard limit enforced by Git, usually lines in a commit's description are wrapped at 72 characters, so that it is easier to read the text when using `git` from the command line. Could you please do so for the commits in this PR?

BrunoRosendo:
> Looks good! After seeing a few tests for different commands, I think that it was a very good idea to implement and use the `testCmdRun` function!
> 
> In any case, I have seen that sometimes the commit messages are pretty long. Even though it is not a hard limit enforced by Git, usually lines in a commit's description are wrapped at 72 characters, so that it is easier to read the text when using `git` from the command line. Could you please do so for the commits in this PR?

I tried to reduce the commit message

tiborsimko (approved):

PR #79, opened 2022-08-15
tests: unit tests for ls and logs commands

### LINKED ISSUE #70: cli: create tests for existing functionalities and commands

BrunoRosendo:
closes #70 

review after #78

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -72,6 +72,9 @@ type logsOptions struct {
 	size       int64
 }
 
+var logsSingleFilters = []string{"compute_backend", "docker_img", "status"}
+var logsMultiFilters = []string{"step"}
+
 // newLogsCmd creates a command to get workflow logs.
 func newLogsCmd() *cobra.Command {
 	o := &logsOptions{}
@@ -155,13 +158,9 @@ func (o *logsOptions) run(cmd *cobra.Command) error {
 }
 
 // parseLogsFilters parses a list of filters in the format 'filter=value', for the 'logs' command.
-// The steps filters are returned as a slice of strings, while the other filters are returned as a map.
 // Returns an error if any of the given filters are not valid.
 func parseLogsFilters(filterInput []string) (utils.Filters, error) {
-	singleValueFilters := []string{"compute_backend", "docker_img", "status"}
-	multiValueFilters := []string{"step"}
-
-	filters, err := utils.NewFilters(singleValueFilters, multiValueFilters, filterInput)
+	filters, err := utils.NewFilters(logsSingleFilters, logsMultiFilters, filterInput)
 	if err != nil {
 		return filters, err
 	}
--- a/cmd/logs_test.go
+++ b/cmd/logs_test.go
@@ -1,0 +1,308 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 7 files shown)


cmd/logs_test.go
```diff
@@ -0,0 +1,306 @@
...
+				}
+			},
+			"engine_specific": "engine logs"
+		}`
+	successResponseRaw, _ := json.Marshal(operations.GetWorkflowLogsOKBody{
+		Logs:         fmt.Sprintf(logsTemplate, "workflow_1", "job1", "workflow 1 logs"),
+		User:         "user",
+		WorkflowID:   "my_workflow_id",
+		WorkflowName: "my_workflow",
+	})
```
BrunoRosendo, line 60: The response was getting a mess so I decided to marshal the body struct instead of writing a string. Should I do this in other places too?

tiborsimko: Do you mean in `logs` or everywhere else? Hmm, seeing textual "success responses" for commands like `info` in the test file actually feels nice to me; we can see IRL about this change.

BrunoRosendo: I meant everywhere else. If you think it's better as text, then we can leave it like that

cmd/logs.go
```diff
@@ -72,6 +72,9 @@ type logsOptions struct {
 	size       int64
 }
 
+var logsSingleFilters = []string{"compute_backend", "docker_img", "status"}
+var logsMultiFilters = []string{"step"}
```
tiborsimko, line 76: The fact that some `logs` command filters are single-natured and some multi-natured is a bit hidden here.  Perhaps we can move this REANA system knowledge outside into some "configuration" file?  See also #80. Something for later?

BrunoRosendo: I agree, it'd be useful to move some constants like these to a config file, it probably makes sense to do it in #80 

The single or multi nature of the filters also feel a bit weird to me, but that comes from the Python Client. In this example, why do we not let the users choose multiple status or compute_backend? This doesn't seem to be in the help message either

tiborsimko: Yeah, it's not currently consistent in the Python client.  For example:

- `reana-client logs -w root6-roofit-yadage-kubernetes --filter status=pending --filter status=failed`
- `reana-client logs -w root6-roofit-yadage-kubernetes --filter step=gendata --filter step=fitdata`

The original thinking behind filtering options was the following: you get results, and you filter what you want to narrow to, and _then_ you filer on the filtered results what you want to narrow ever more. In this optics, both of the above commands should return nothing.  But, currently, whilst the first command conforms to this logic, the latter command returns both steps instead of nothing.  So the analogy with Unix pipe filters is retained for the "status" filter, but broken for the "step" filter, one could say.

P.S. The obvious usage would be something like `--filter name=roofit --filter status=finished --filter step=gendata`. But the above commands help to think about joining up filters more easily.

We can ticketise this for later, and either improve the documentation, or change the behaviour.

BrunoRosendo: Sounds good, I think we can merge this then

tiborsimko (approved):
LGTM. We may want to refactor some configuration-level parts as part of #80 later.

PR #82, opened 2022-08-17
cli: implement `diff` command

### LINKED ISSUE #32: cli: implement `diff` command
Create a new `diff` command that replicates all the functionality we have in `reana-client`.
It should also include all the flags and return exactly the same output as `reana-client diff`.

Please note that this command heavily uses coloring, so it would be better to implement https://github.com/reanahub/reana-client-go/issues/2 issue before this one.

Example:
```
$ reana-client diff workflow.1 workflow.2
==> Differences in workflow workspace
@@ -1 +1 @@

-{'retention_days': {'test/*.root': 5}}
+{'retention_days': {'**/*': 5, 'test/*.root': 10}}

==> Differences in workflow workspace
Binary files workflow.1/results/data.root and workflow.2/results/data.root differ

```

BrunoRosendo:
closes #32

### DIFF
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -1,0 +1,174 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/utils"
+
+	"github.com/jedib0t/go-pretty/v6/text"
+
+	"github.com/spf13/cobra"
+)
+
+const diffDesc = `
+Show diff between two workflows.
+
+The ` + "``diff``" + ` command allows to compare two workflows, the workflow_a and
+workflow_b, which must be provided as arguments. The output will show the
+difference in workflow run parameters, the generated files, the logs, etc.
+
+Examples:
+
+	$ reana-client diff myanalysis.42 myotheranalysis.43
+
+	$ reana-client diff myanalysis.42 myotheranalysis.43 --brief
+`
+
+type diffOptions struct {
+	token     string
+	workflowA string
+	workflowB string
+	brief     bool
+	unified   int
+}
+
+// newDiffCmd creates a command to show diff between two workflows.
+func newDiffCmd() *cobra.Command {
+	o := &diffOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "diff",
+		Short: "Show diff between two workflows.",
+		Long:  diffDesc,
+		Args:  cobra.ExactArgs(2),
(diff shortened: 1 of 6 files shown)


cmd/diff.go
```diff
@@ -0,0 +1,174 @@
...
+		if hasWorkflow {
+			specificationDiff["specification"] = val
+			delete(specificationDiff, "workflow")
+		}
+		equalSpecification := true
+		for section, lines := range specificationDiff {
+			if len(lines) != 0 {
+				equalSpecification = false
+				utils.PrintColorable(
+					fmt.Sprintf("%s Differences in workflow %s\n", leadingMark, section),
```
tiborsimko, line 120: One thing I noticed with respect to the Python client is the difference section ordering:

```console
$ reana-client diff events.1 test | grep "^==>"
==> Differences in workflow inputs
==> Differences in workflow outputs
==> Differences in workflow version
==> Differences in workflow workspace
==> Differences in workflow specification
==> Differences in workflow workspace

$ $ ./reana-client-go diff events.1 test | grep "==>"
==> Differences in workflow specification
==> Differences in workflow inputs
==> Differences in workflow outputs
==> Differences in workflow version
==> Differences in workflow workspace
==> Differences in workflow workspace
```

Some observations:

- It's weird that there are two "workflow workspace" sections, but that's also true for the Python client, and should be fixed there. So nothing to do here.
- Can we align the output order to respect Python client?

BrunoRosendo: The ordering is different because GO's maps are unordered, unlike in Python. It's possible to maintain order by using [json.Decoder](https://pkg.go.dev/encoding/json#Decoder) instead of the [json.Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) being used. The code will probably be a little bit more verbose but the change shouldn't be too difficult

tiborsimko: Shall we do it now or after package refactoring?

BrunoRosendo: We can do it now

BrunoRosendo: Turns out I was wrong about using the Decoder, it works nicely for arrays but not so much for objects, since they don't guarantee order by nature. By looking at [this discussion](https://github.com/golang/go/issues/27179) I found some hacky workarounds but no nice solution for this, the developers don't seem to support it because for order one should use an array.

An example of not so nice solution, is using the Decoder and a manually parse of the JSON, it doesn't look good at all: https://go.dev/play/p/jOXpSDpJz04

BrunoRosendo: I'm not sure if it's worth implementing this order (if we want order, shouldn't the server return it as an array?)

tiborsimko: Yeah, fully agreed that it would be very clean to do this on the server-side. But since it would necessitate breaking REST API response changes, it's not something for now.

Hmm, looking at the pasted workaround, it might do actually as a stop-gap solution; however it's also possible that we can live with the Python client <-> Go client output differences for this commend, since it is not probable that some clients has wrapped it into some wider scripted glue usage scenarios... I'm kind of 60:40 split on this.

tiborsimko (approved):
Works nicely :+1: The output ordering issue to be ticketised for later

PR #83, opened 2022-08-17
global: re-organize package structure

### LINKED ISSUE #80: RFC packages, utilities, and file structure
**Current behaviour**

This issue is an RFC about package structure that starts to grow as we are adding more and more functionality to the client.

We currently have the following utility file structure:

- `utils/data.go`
- `utils/display.go`
- `utils/filters.go`
- `utils/utils.go`
- `validation/utils.go`

One can ask why "validation utilities" went into its own package `validation/utils.go` whilst "filtering utilities" went into `utils/filters.go`? In other words, when does one promote a utility into its own package, and why is there a "utils utilities" `utils/utils.go` at all?

It would be nice to have a schema for components so that the role of each package and file in the overall repository structure would be very clear just from browsing the file system.

**Expected behaviour**

I think we should try to harmonise the presence of various "utilities" as packages/files go, for example:

A. One option is to prefer having one global utility package (`utils`) that would group all topical utilities such as filters under its umbrella. This would call for `validation/utils.go` to be moved to `utils/validation.go` perhaps?

B. Another option is to prefer having topical utilities living as separate entities whenever possible. This would call for keeping `validation/utils.go` and moving `utils/filters.go` into say `filtering` package, having perhaps `filtering/utils.go` or `filtering/filters.go` inside?

C. Note also that the purpose of the generically named "utility utilities" (`utils/utils.py`) can be pretty opaque. We could separate out functions from generic `utils/util.go` into their own specific and more descriptive topical utilities. Some functions are of "configuration" nature describing REANA defaults; these we usually have in `config.py` in Python packages, so we could do something similar for consistency. Other functions are "errors handlers" in nature, they could merit their own file (or perhaps even package); some are "time stamp formatters" so they could be associated with the other formatters in the same package already, etc.

Personally, I always tend to think about a system from the point of view of associating components represented by nouns; e.g. there is "formatter", there is "validator", there is "filterer", there is "displayer", there is "error handler", there could be ""downloader", "verifier", etc. And perhaps only one small generic utility catch-all component only.

The goal of this issue is to design a nice package/file structure for the Go client that would help to (1) fix current decision fatigue regarding `validation/utils.go` vs `utils/validation.go` when adding a new utility; (2) expose better the various topical utilities we have in the system; as well as (3) allow the client to organically and logically grow as we shall be progressively adding new R/W functionalities regarding uploading downloading files, secrets, etc.

BrunoRosendo:
closes #80 

review after #82


pkg/data_utils/data_utils.go
```diff
@@ -0,0 +1,58 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+// Package data_utils gives extra functions to manipulate data structures like strings and slices.
```
tiborsimko, line 9: Cosmetics: Please rename `data_utils` to `datautils` or `datautil` everywhere, i.e. without underscore. Let's try to keep the package name to one word only.  Ditto for `error_handler` -> `errorhandler`.

pkg/workflows/workflows.go
```diff
@@ -0,0 +1,49 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+// Package workflows gives utility functions related to REANA's workflows.
+package workflows
```
tiborsimko, line 10: The package name `workflows` seems "important", but its contents seems "underused". Might be better called `workflowutils`, but that's weird.  We can keep it as it, and think whether it wouldn't grow in importance, say `workflows/list.go`, `workflows/status.go`, etc to hold more of the "core" business logic?

config/config.go
```diff
@@ -0,0 +1,60 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+// Package config gives constants and small functions that specify the REANA client configuration.
+package config
```
tiborsimko, line 10: We now have three package levels in this PR: (i) `cmd` with cobra stuff, (ii) `config` for abstracting hard-coded values, and (iii) `pkg` where all the displaying, formatting etc utility live (in their respective packages). The fact that a lot of business logic lives under `cmd` and not `pkg` creates a certain dissonance.

It may be good to find a way to have `cmd` stored under `pkg`, as other software projects do (e.g. kubectl, kind). In this way it'll be clear where the most important stuff is.  Also, `config` should then be moved there.

Note also that kubectl, kind and friends usually not use single `cmd` package for everything, but introduce packages for each command, which in our case would be `list`, `status`, etc. This would make sense because the whole reana-client-go is about CLI commands, so keeping cobra layer tiny, just for parsing CLI options etc, whilst having clearly visible business logic packages implementing each command business logic, would makes sense. From the pure package naming point of view, it seems advantageous to keep "cmd" tiny and place focus rather on the business logic itself.

Something we can think for later as the functionality grows.

pkg/workflows/workflows_test.go
```diff
@@ -0,0 +1,107 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package workflows
```
tiborsimko, line 9: BTW with the changed code organisation structure, we can rewrite the `Makefile` code coverage testing command from:

```
go test -cover `go list ./... | grep -v /reana-client-go/client`
```

to use the three leading leading directories that we have introduced:

```
go test -cover ./cmd/... ./config/... ./pkg/...
```

pkg/workflows/workflows.go
BrunoRosendo: If we're planning to extend the package that way, should we at least rename the file to `utils.go`, without changing the package name?

config/config.go
BrunoRosendo: I moved `config` inside `pkg`. Separating the business logic from the cobra template seems out of reach of this PR but I could try creating different packages for each command. What do you think?

tiborsimko: Yes, we can go like this, and think about possible further changes as we shall be adding upload/secret/etc functionality.

tiborsimko (approved):
Works nicely, and looks better :+1: And we can consider some further changes to the "core" business logic, such as using single `cmd` package for all commands, vs using separate packages for separate commands or command groups, later on.

PR #89, opened 2022-08-22
tests: implement RemoveFromSlice unit tests

### LINKED ISSUE #88: datautils: `RemoveFromSlice` missing unit tests
The `RemoveFromSlice` function inside `datautils` package is currently missing unit tests. We should write them, the same way it's done for the other functions in the same package

BrunoRosendo:
closes #88

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ swagger-validate-specs: # Validate OpenAPI specification.
 	$(SWAGGER) validate "../reana-server/docs/openapi.json"
 
 test: # Run test suite.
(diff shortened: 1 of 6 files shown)


tiborsimko (approved):
Nice :+1: 

I'll get back to `Makefile`. One thing that slipped away whilst merging #83 is the removal of subcommand; instead of:

```
go test -cover `go list ./cmd/... ./pkg/...`
```

we can simply have:

```
go test -cover ./cmd/... ./pkg/...
```

Can you please fix it here?

PR #91, opened 2022-08-23
cli: update info command with new information

### LINKED ISSUE #90: cli: update `info` command with new information
The reana-server was updated with extra information that is returned in the `info` endpoint. To reflect this changes, one needs to update the generated client, from the openapi specification, and print the extra fields to the standard output

BrunoRosendo:
closes #90


tiborsimko (approved):
Works nicely :+1: and better than Python client :wink: https://github.com/reanahub/reana-client/issues/627

PR #103, opened 2022-08-23
cli: implement delete command

### LINKED ISSUE #93: cli: implement `delete` functionality
```console
$ reana-client delete --help
Usage: reana-client delete [OPTIONS]

  Delete a workflow.

  The ``delete`` command removes workflow run(s) from the database. Note that
  the workspace will always be deleted, even when ``--include-workspace`` is
  not specified. Note also that you can remove all past runs of a workflow by
  specifying ``--include-all-runs`` flag.

  Example:

       $ reana-client delete -w myanalysis.42

       $ reana-client delete -w myanalysis.42 --include-all-runs

Options:
  --include-all-runs       Delete all runs of a given workflow.
  --include-workspace      Delete workspace from REANA.
```

BrunoRosendo:
closes #93

### DIFF
--- a/pkg/workflows/manager.go
+++ b/pkg/workflows/manager.go
@@ -1,0 +1,46 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package workflows
+
+import (
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/pkg/config"
+	"reanahub/reana-client-go/pkg/validator"
+)
+
+// UpdateStatus updates the status of the specified workflow.
+func UpdateStatus(
+	token, workflow, status string,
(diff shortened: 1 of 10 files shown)


pkg/workflows/manager.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = api.Operations.SetWorkflowStatus(deleteParams)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
BrunoRosendo, line 46: I decided to put this function here because it can be used by other commands. However, the test coverage is decreased, since this is only tested in `TestDelete` in another package

tiborsimko (commented):
Works well :+1: 

Left one comment about documentation.

BTW if one deletes an already deleted workflow, then the output message may be cryptic:

```console
$ reana-client delete -w demo
==> SUCCESS: demo has been deleted
$ reana-client delete -w demo
==> SUCCESS: demo has been deleted
```

One would expect the second message to read `ERROR: demo was already deleted`. But the Python client does the same, so we can address this observation later for both.

PR #107, opened 2022-08-25
cli: implement secrets-list command

### LINKED ISSUE #98: cli: implement `secrets-list` functionality
```console
$ reana-client secrets-list --help
Usage: reana-client secrets-list [OPTIONS]

  List user secrets.

  Examples:

       $ reana-client secrets-list

Options:
```

Note that there will be also a `secrets-add` and `secrets-delete` commands. So this could serve as an example of a Go package `secrets` that exposes several CLI commands.  (But we can keep using `cmd` still.)

BrunoRosendo:
closes #98

### DIFF
--- a/cmd/secrets_list.go
+++ b/cmd/secrets_list.go
@@ -1,0 +1,73 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/pkg/displayer"
+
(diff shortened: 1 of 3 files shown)


cmd/secrets_list.go
```diff
@@ -0,0 +1,74 @@
...
+	listSecretsResp, err := api.Operations.GetSecrets(listSecretsParams)
+	if err != nil {
+		return err
+	}
+
+	header := []string{"name", "type"}
+	var rows [][]any
+	for _, secret := range listSecretsResp.Payload {
+		var row []any
+		row = append(row, secret.Name, secret.Type)
```
mdonadoni, line 68:
```suggestion
	var rows [][]string
	for _, secret := range listSecretsResp.Payload {
		row := []string{secret.Name, secret.Type}
```
WDYT?

BrunoRosendo: Much better, thank you

mdonadoni (approved):

PR #108, opened 2022-08-25
cli: implement rm command

### LINKED ISSUE #96: cli: implement `rm` functionality
```console
$ reana-client rm --help
Usage: reana-client rm [OPTIONS] SOURCES

  Delete files from workspace.

  The ``rm`` command allow to delete files and directories from workspace.
  Note that you can use glob to remove similar files.

  Examples:

       $ reana-client rm -w myanalysis.42 data/mydata.csv

       $ reana-client rm -w myanalysis.42 'data/*root*'

Options:
```

BrunoRosendo:
closes #96

### DIFF
--- a/cmd/rm.go
+++ b/cmd/rm.go
@@ -1,0 +1,129 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/pkg/config"
+	"reanahub/reana-client-go/pkg/displayer"
+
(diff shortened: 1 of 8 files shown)


cmd/rm.go
```diff
@@ -0,0 +1,117 @@
...
+				false,
+				cmd.OutOrStdout(),
+			)
+		}
+		for file, errorInfo := range failed {
+			displayer.DisplayMessage(
+				fmt.Sprintf("Something went wrong while deleting %s.\n%s", file, errorInfo.Error),
+				displayer.Error,
+				false,
+				cmd.OutOrStdout(),
```
BrunoRosendo, line 104: This means the client will exit on a 0 code even if there's an error like this one. Is this a problem?

It's hard to recreate Python's client of displaying all the error and then doing `sys.exit(1)`. We could do that here but we're using the flow of returning an error to the cobra's calling function and this would break it

BrunoRosendo: Handled by returning an empty error and exiting 1 without displaying it

mdonadoni (approved):
Looks good to me!

PR #111, opened 2022-08-26
cli: implement mv command

### LINKED ISSUE #97: cli: implement `mv` functionality
```console
$ reana-client mv --help
Usage: reana-client mv [OPTIONS] SOURCE TARGET

  Move files within workspace.

  The ``mv`` command allow to move the files within workspace.

  Examples:

       $ reana-client mv data/input.txt input/input.txt

Options:
```

BrunoRosendo:
closes #97

### DIFF
--- a/cmd/mv.go
+++ b/cmd/mv.go
@@ -1,0 +1,89 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/pkg/displayer"
+
+	"github.com/spf13/cobra"
+)
+
+const mvDesc = `
+Move files within workspace.
+
+The ` + "``mv``" + ` command allows to move the files within workspace.
+
+Examples:
+
+	$ reana-client mv data/input.txt input/input.txt
+`
+
+type mvOptions struct {
+	token    string
+	workflow string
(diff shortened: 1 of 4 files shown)


cmd/mv.go
```diff
@@ -0,0 +1,116 @@
...
+	sourceExists := false
+	for _, file := range files.Items {
+		if strings.Contains(file.Name, o.source) {
+			sourceExists = true
+			break
+		}
+	}
+	if !sourceExists {
+		return fmt.Errorf("source file(s) %s does not exist in %s workspace", o.source, o.workflow)
+	}
```
BrunoRosendo, line 91: Should we keep this validation? I feel like we could just use the validation from the server but I wasn't sure so I ported the way it's done in Python

mdonadoni: Yes, I think it is a good idea to remove both of them, as they are already being done server side. However, the server side check that a workflow is not running is not working correctly, so I will open an issue about that.

mdonadoni: Related issue: https://github.com/reanahub/reana-workflow-controller/issues/469

BrunoRosendo: It's done, I had to update the openapi and regenerate the client

mdonadoni (approved):

cmd/mv.go
```diff
@@ -0,0 +1,89 @@
...
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/pkg/displayer"
+
+	"github.com/spf13/cobra"
+)
+
+const mvDesc = `
+Move files within workspace.
+
+The ` + "``close``" + ` command allows to move the files within workspace.
```
mdonadoni, line 23:
```suggestion
The ` + "``mv``" + ` command allows to move the files within workspace.
```

PR #114, opened 2022-08-30
cli: implement secrets-add command

### LINKED ISSUE #100: cli: implement `secrets-add` functionality
```console
$ reana-client secrets-add --help
Usage: reana-client secrets-add [OPTIONS]

  Add secrets from literal string or from file.

  Examples:

       $ reana-client secrets-add --env PASSWORD=password

       $ reana-client secrets-add --file ~/.keytab

       $ reana-client secrets-add --env USER=reanauser

                                  --env PASSWORD=password

                                  --file ~/.keytab

Options:
  --env TEXT               Secrets to be uploaded from literal string.e.g.
                           PASSWORD=password123
  --file FILE              Secrets to be uploaded from file.
  --overwrite              Overwrite the secret if already present
```

Note that this will require to implement (tiny) file uploader, so this may be better kept for later after #98 and #99 are done. We can use standard Go libraries for the uploader concept here, and attack `upload` and `download` commands at the same time as this issue.

BrunoRosendo:
closes #100

### DIFF
--- a/cmd/secrets_add.go
+++ b/cmd/secrets_add.go
@@ -1,0 +1,158 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"encoding/base64"
+	"fmt"
+	"os"
+	"path/filepath"
+	"reanahub/reana-client-go/client"
+	"reanahub/reana-client-go/client/operations"
+	"reanahub/reana-client-go/pkg/datautils"
+	"reanahub/reana-client-go/pkg/displayer"
+	"reanahub/reana-client-go/pkg/validator"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+const secretsAddDesc = `
+Add secrets from literal string or from file.
+
+Examples:
+
+	$ reana-client secrets-add --env PASSWORD=password
+
+	$ reana-client secrets-add --file ~/.keytab
+
+	$ reana-client secrets-add --env USER=reanauser
+
+				   --env PASSWORD=password
+
+				   --file ~/.keytab
+`
+
+type secretsAddOptions struct {
+	token       string
+	envSecrets  []string
+	fileSecrets []string
+	overwrite   bool
+}
+
+// newSecretsAddCmd creates a command to add secrets from literal string or from file.
+func newSecretsAddCmd() *cobra.Command {
+	o := &secretsAddOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "secrets-add",
+		Short: "Add secrets from literal string or from file.",
+		Long:  secretsAddDesc,
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := validator.ValidateAtLeastOne(
+				cmd.Flags(), []string{"env", "file"},
+			); err != nil {
+				return fmt.Errorf("%s\n%s", err.Error(), cmd.UsageString())
+			}
+			for _, file := range o.fileSecrets {
+				if err := validator.ValidateFile(file); err != nil {
+					return fmt.Errorf("invalid value for '--file': %s", err.Error())
+				}
+			}
+			return o.run(cmd)
+		},
+	}
+
+	f := cmd.Flags()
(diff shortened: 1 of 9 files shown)


cmd/secrets_add.go
```diff
@@ -0,0 +1,158 @@
...
+			if err := validator.ValidateAtLeastOne(
+				cmd.Flags(), []string{"env", "file"},
+			); err != nil {
+				return fmt.Errorf("%s\n%s", err.Error(), cmd.UsageString())
+			}
+			for _, file := range o.fileSecrets {
+				if err := validator.ValidateFile(file); err != nil {
+					return fmt.Errorf("invalid value for '--file': %s", err.Error())
+				}
+			}
```
BrunoRosendo, line 68: In order to be consistent with the Python client, this validation is done before actually reading and uploading the files. However, they could, for some reason, stop existing or have their permissions changed before uploading them. This is a concurrency edge case and, if it happens, we return the error given by `ReadFile` instead of the pretty ones from the validation

cmd/secrets_add.go
```diff
@@ -0,0 +1,158 @@
...
+
+	for _, filePath := range fileSecrets {
+		data, err := os.ReadFile(filePath)
+		if err != nil {
+			return nil, nil, fmt.Errorf(
+				"file %s could not be uploaded: %s",
+				filePath, err.Error(),
+			)
+		}
+		encodedData := base64.StdEncoding.EncodeToString(data)
```
BrunoRosendo, line 147: We talked about creating a tiny uploader package, but the "upload" itself can be made in 2 lines, given that the encoding is specific to `secrets`, so I just did it here.

mdonadoni (approved):
Looks good! I also tested this PR by uploading the secrets necessary for Kerberos and everything works as expected.

pkg/validator/validator_test.go
```diff
@@ -216,6 +221,57 @@ func TestValidateOperationalOptions(t *testing.T) {
...
+		},
+	}
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			if test.notReadable {
+				err := os.Chmod(test.path, 0222)
+				if err != nil {
+					t.Fatalf("Failed to change permissions of file %s: %s", test.path, err.Error())
+				}
+			}
```
mdonadoni, line 258: This causes some troubles with git, for example checking out another branch fails:
```console
% git checkout master
error: Your local changes to the following files would be overwritten by checkout:
	testdata/inputs/noperms.txt
Please commit your changes or stash them before you switch branches.
Aborting
```
What do you think about setting these permissions directly on the file, given Git should preserve them? In this way, there is also no need to the set them from the test.

mdonadoni, line 258: This causes some troubles with git, for example checking out another branch fails:
```console
% git checkout master
error: Your local changes to the following files would be overwritten by checkout:
	testdata/inputs/noperms.txt
Please commit your changes or stash them before you switch branches.
Aborting
```
What do you think about setting these permissions directly on the file, given Git should preserve them?

BrunoRosendo: That was my initial solution but, for some reason, the Github Actions CI wasn't preserving them (I though it could be something like root perms when running the tests but changing them in the code fixed it). I can try to find another way, since this approach doesn't work well either

BrunoRosendo: It seems like git actually doesn't preserve them upon checkout and it won't keep them for you: https://stackoverflow.com/questions/14557106/git-changes-my-file-permissions-upon-checkout

It seems we have to do it manually or remove this test. Any thoughts?

BrunoRosendo: Fixed by creating a temporary file instead

cmd/secrets_add_test.go
```diff
@@ -0,0 +1,196 @@
...
+	tempDir := t.TempDir()
+	emptyFile := tempDir + "/empty.txt"
+	piFile := tempDir + "/pi.txt"
+	_, err := os.Create(emptyFile)
+	if err != nil {
+		t.Fatalf("Error while creating empty file: %s", err.Error())
+	}
+	err = os.WriteFile(piFile, []byte("3.14"), 0777)
+	if err != nil {
+		t.Fatalf("Error while creating noperms file: %s", err.Error())
```
mdonadoni, line 98:
```suggestion
		t.Fatalf("Error while creating pi.txt file: %s", err.Error())
```

Issue #17, opened 2022-07-07
cli: group commands in help message


audrium:
In `reana-client` we group certain commands into sections, e.g. `Configuration commands`, `Workflow management commands`, etc.
Cobra does not yet provide this functionality out of the box. Though there is already a PR prepared with this feature and merging this PR is included into the `Summer 2022 release` plan.. So perhaps we could simply wait until it will be officially released.
Alternative would be to investigate how others are providing a custom template for Cobra: [docker](https://github.com/docker/cli/blob/4460252b43bd77f237a54c46ffdfe3c73b52f23f/cli/cobra.go#L406), [kubectl](https://github.com/kubernetes/kubectl/blob/ca3557e2c6a00b47823c07b1b51944cba45b5c92/pkg/cmd/cmd.go#L327)

PR #123, opened 2022-09-13
cli: group commands in help message

audrium:
closes https://github.com/reanahub/reana-client-go/issues/17

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -11,6 +11,7 @@ package cmd
 
 import (
 	"os"
+	"reanahub/reana-client-go/pkg/commandgroups"
 	"reanahub/reana-client-go/pkg/validator"
 
 	"github.com/spf13/pflag"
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -11,6 +11,7 @@ package cmd
 
 import (
 	"os"
+	"reanahub/reana-client-go/pkg/templater"
```
tiborsimko, line 14: Hmm, `templater` seems rather generic, as it could refer to our own outputs. What about naming the package like `cobratemplates` or `cobracommands` or `cobragroups` or `cligroups` or `commandgroups` or something of that kind?

pkg/templater/templater.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+			groups = append(groups, strings.Join(cmds, "\n"))
+		}
+		return strings.Join(groups, "\n\n")
+	})
+	c.SetUsageTemplate(usageTemplate)
+}
+
+// rpad adds padding to the right of a string.
+func rpad(s string, padding int) string {
```
tiborsimko, line 52: Can you add some basic test case for `rpad()` (or more) in order to have at least _some_ code coverage for the new package?

```console
$ make test | grep templater
?       reanahub/reana-client-go/pkg/templater  [no test files]
```

cmd/root.go
```diff
@@ -46,26 +47,71 @@ func NewRootCmd() *cobra.Command {
...
+				// run
+				// stop
+				// validate
+				newLogsCmd(),
+				newStartCmd(),
+				newStatusCmd(),
+			},
+		},
+		{
+			Message: "Workspace interactive commands:",
```
tiborsimko, line 87: Looks nice and works well :+1:

cmd/root.go
audrium: thanks, I like `commandgroups` the most

tiborsimko (approved):

Issue #101, opened 2022-08-23
profile `list` performance


tiborsimko:
After fixing `list` command performance in #63 and #92, the Python vs Go client speed advantage is back to normal.

However, whlist for smaller outputs Go can be 5x faster than Python:

```console
$ hyperfine --warmup 5 'reana-client list --size 20' './reana-client-go list --size 20'
Benchmark 1: reana-client list --size 20
  Time (mean ± σ):     628.7 ms ±   9.5 ms    [User: 464.2 ms, System: 42.7 ms]
  Range (min … max):   617.9 ms … 643.6 ms    10 runs

Benchmark 2: ./reana-client-go list --size 20
  Time (mean ± σ):     133.3 ms ±  19.1 ms    [User: 6.0 ms, System: 6.4 ms]
  Range (min … max):   118.8 ms … 190.7 ms    23 runs

Summary
  './reana-client-go list --size 20' ran
    4.72 ± 0.68 times faster than 'reana-client list --size 20'
```

for larger outputs the difference can be less than 2x faster only:

```console
$ ./reana-client-go list | wc -l
824
$ hyperfine --warmup 5 'reana-client list' './reana-client-go list'
Benchmark 1: reana-client list
  Time (mean ± σ):      1.254 s ±  0.030 s    [User: 0.528 s, System: 0.055 s]
  Range (min … max):    1.199 s …  1.287 s    10 runs

Benchmark 2: ./reana-client-go list
  Time (mean ± σ):     691.5 ms ±  47.1 ms    [User: 20.6 ms, System: 12.1 ms]
  Range (min … max):   639.8 ms … 781.6 ms    10 runs

Summary
  './reana-client-go list' ran
    1.81 ± 0.13 times faster than 'reana-client list'
```

This is perhaps simply due to longer REST API response transfer times, but perhaps we should profile where the most time is spent in the Go client when handling when the `list` output, and see whether we could improve anything here.

BTW we may perhaps take this occasion to revive the idea of introducing a `--profile` option to commands like `list` in order make these measurements easily doable and also repeatable in the future.

audrium:
With https://github.com/reanahub/reana-client-go/pull/124 we introduced a generic profiler to all cli commands. It produces a profiling data file which could be analyzed with [pprof](https://github.com/google/pprof). To install it do: `go get -u github.com/google/pprof`. Usage example:
```
go build
./reana-client-go list --profile=cpu
go tool pprof -text profile-cpu.pprof
```
[...]

PR #124, opened 2022-09-15
cli: add generic profiler

audrium:
closes https://github.com/reanahub/reana-client-go/issues/101

to test:
```
go run . list --profile=cpu
go tool pprof -text profile.pprof
```

### DIFF
--- a/cmd/profiling.go
+++ b/cmd/profiling.go
@@ -1,0 +1,87 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"runtime"
+	"runtime/pprof"
+
+	"github.com/spf13/pflag"
+)
+
+var (
+	profileMode         string
+	profileOutputFormat string = "profile-%s.pprof"
+)
+
+func addProfilerFlags(flags *pflag.FlagSet) {
+	flags.StringVar(
+		&profileMode,
+		"profile",
+		"none",
+		"Enable profiling. One of (none|cpu|heap)",
+	)
+}
+
+func setupProfiler() error {
+	var (
+		f   *os.File
+		err error
+	)
+	switch profileMode {
+	case "none":
+		return nil
+	case "cpu":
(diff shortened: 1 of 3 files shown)


cmd/profiling.go
```diff
@@ -0,0 +1,102 @@
...
+		&profileMode,
+		"profile",
+		"none",
+		"Enable profiling. One of (none|cpu|heap)",
+	)
+	flags.StringVar(
+		&profileOutput,
+		"profile-output",
+		"profile.pprof",
+		"Name of the file to write the profile to",
```
audrium, line 37: Not sure if we need a separate flag to control the output file, since we print all the persistent flags in help message (`go run .`) it kind of pollutes the output. We could remove it to keep it simple

tiborsimko: Perhaps we could reuse the `--profile` option value?  E.g. if `--profile=cpu`, then we cloud create `profile-cpu.pprof`, and ditto for `--profile=heap`, etc.

cmd/profiling.go
```diff
@@ -0,0 +1,91 @@
...
+	)
+	switch profileMode {
+	case "none":
+		return nil
+	case "cpu":
+		f, err = os.Create(profileOutput)
+		if err != nil {
+			return err
+		}
+		err = pprof.StartCPUProfile(f)
```
tiborsimko, line 53: Works well :+1: Although the sample coverage does not look great, as we discussed IRL. E.g. for one of my tests:

```
Duration: 1.06s, Total samples = 30ms ( 2.83%)
Showing nodes accounting for 30ms, 100% of 30ms total
      flat  flat%   sum%        cum   cum%
      10ms 33.33% 33.33%       10ms 33.33%  encoding/json.(*decodeState).rescanLiteral
      10ms 33.33% 66.67%       10ms 33.33%  encoding/json.stateInString
      10ms 33.33%   100%       10ms 33.33%  runtime.(*spanSet).push
         0     0%   100%       20ms 66.67%  encoding/json.(*Decoder).Decode
         0     0%   100%       10ms 33.33%  encoding/json.(*Decoder).readValue
         0     0%   100%       10ms 33.33%  encoding/json.(*decodeState).array
...
```

If parametrising sample taking via a new CLI option wouldn't help much here, then we could merge "as is"

tiborsimko (approved):

Issue #122, opened 2022-09-06
cli: implement all the write commands


audrium:
This is an epic umbrella issue to hold a list of all the write commands which needs to be implemented in the go client:

- [ ] `create` 
- [x] `delete` https://github.com/reanahub/reana-client-go/issues/93
- [x] `restart` https://github.com/reanahub/reana-client-go/pull/125
- [ ] `run`
- [x] `start` https://github.com/reanahub/reana-client-go/issues/95
- [x] `stop` https://github.com/reanahub/reana-client-go/pull/127
- [x] `mv`https://github.com/reanahub/reana-client-go/issues/97
- [x] `rm` https://github.com/reanahub/reana-client-go/issues/96
- [ ] `download`
- [ ] `upload`
- [x] secrets-* commands: add, delete, list

audrium:
Closing since most of the commands are already implemented and the remaining ones have a dedicated issues

PR #125, opened 2022-09-19
cli: implement restart command

audrium:
addresses https://github.com/reanahub/reana-client-go/issues/122

Please note that this PR misses implementation of ``--file`` flag, because it requires [loading and validating](https://github.com/reanahub/reana-client/blob/master/reana_client/cli/workflow.py#L601-L604) the specification file. It could be tackled later once we make the clients thin and refactor the spec upload logic to send the whole yaml file to the server side and process it there.

### DIFF
--- a/cmd/restart.go
+++ b/cmd/restart.go
@@ -1,0 +1,165 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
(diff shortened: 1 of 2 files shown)


tiborsimko (approved):
Works nicely :+1: Tested also parameter passing or status checking.

Issue #131, opened 2022-09-21
cli: implement `upload` functionality 


audrium:
```
Usage: reana-client upload [OPTIONS] SOURCES

  Upload files and directories to workspace.

  The ``upload`` command allows to upload workflow input files and
  directories. The SOURCES argument can be repeated and specifies which files
  and directories are to be uploaded, see examples below. The default
  behaviour is to upload all input files and directories specified in the
  reana.yaml file.

  Examples:

       $ reana-client upload -w myanalysis.42

       $ reana-client upload -w myanalysis.42 code/mycode.py

Options:
  -w, --workflow TEXT      Name or UUID of the workflow. Overrides value of
                           REANA_WORKON environment variable.
  -t, --access-token TEXT  Access token of the current user.
  --help                   Show this message and exit.
```
Note that initially we can use standard Go libraries for the file upload, and implement advanced upload features using `rclone` later in different issues.

PR #133, opened 2022-09-23
cli: implement upload command

audrium:
closes https://github.com/reanahub/reana-client-go/issues/131

To test:
```
$ go build
$ cd ../reana-demo-helloworld 
$ reana-client create -w upload-test
$ ../reana-client-go/reana-client-go upload -w upload-test reana.yaml -l DEBUG # check https://localhost:30443 for the reana.yaml file to be present in the workspace
$ ../reana-client-go/reana-client-go upload -w upload-test -l DEBUG # check https://localhost:30443 for all the workspace input files to be present in the workspace
# modify reana.yaml to specify some input directories and test that all files are uploaded
```

Please note that https://github.com/reanahub/reana-client-go/issues/134 will be tackled in a separate PR

### DIFF
--- a/cmd/upload.go
+++ b/cmd/upload.go
@@ -1,0 +1,168 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"reanahub/reana-client-go/pkg/displayer"
+	"reanahub/reana-client-go/pkg/workflows"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
+)
+
+const uploadDesc = `
+Upload files and directories to workspace.
+
+The ` + "``upload``" + ` command allows to upload workflow input files and
+directories. The SOURCES argument can be repeated and specifies which files
+and directories are to be uploaded, see examples below. The default
+behaviour is to upload all input files and directories specified in the
+reana.yaml file.
+
+Examples:
+
+  $ reana-client upload -w myanalysis.42
+
+  $ reana-client upload -w myanalysis.42 code/mycode.py
+`
+
+type uploadOptions struct {
+	token    string
+	workflow string
+}
+
+// newUploadCmd creates a command to upload files and directories to workspace.
+func newUploadCmd() *cobra.Command {
+	o := &uploadOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "upload",
+		Short: "Upload files and directories to workspace.",
+		Long:  uploadDesc,
+		Args:  cobra.ArbitraryArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.run(cmd, args)
+		},
+	}
+
+	f := cmd.Flags()
+	f.StringVarP(&o.token, "access-token", "t", "", "Access token of the current user.")
+	f.StringVarP(
+		&o.workflow,
+		"workflow",
+		"w", "",
+		"Name or UUID of the workflow. Overrides value of REANA_WORKON environment variable.",
+	)
+
+	return cmd
+}
+
+// TODO: filter files based on .gitignore and .reanaignore
+
+func (o *uploadOptions) run(cmd *cobra.Command, args []string) error {
+	var inputPaths []string
+
+	if len(args) > 0 {
+		// upload files and directories specified in arguments.
+		inputPaths = args
+	} else {
+		// upload all input files and directories specified in the reana.yaml file.
+		spec, err := workflows.GetWorkflowSpecification(o.token, o.workflow)
+		if err != nil {
+			return err
+		}
+		inputFiles := spec.Specification.Inputs.Files
+		inputDirs := spec.Specification.Inputs.Directories
+		if err := o.validateInputs(inputFiles, inputDirs); err != nil {
+			return err
+		}
+		inputPaths = append(inputFiles, inputDirs...)
+	}
+
+	files, err := o.collectFiles(cmd, inputPaths)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		_, err := workflows.UploadFile(o.token, o.workflow, file)
+		if err != nil {
+			return err
+		}
+		displayer.DisplayMessage(
+			fmt.Sprintf("File %s was successfully uploaded.", file),
+			displayer.Success,
+			false,
+			cmd.OutOrStdout(),
+		)
+	}
+	return nil
+}
+
+func (o *uploadOptions) collectFiles(cmd *cobra.Command, inputPaths []string) ([]string, error) {
+	log.Debugf("Traverse all the input paths to collect files which needs to be uploaded")
+	log.Debugf("paths: %s", strings.Join(inputPaths, ", "))
+	var files []string
+	for _, dir := range inputPaths {
+		err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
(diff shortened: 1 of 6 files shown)


mdonadoni:
Can you also reproduce this issue when dealing with symlinks?

Directory structure:
```
test
├── dir
│   └── inside.txt
├── dir_symlink -> dir
├── file_symlink.txt -> outside.txt
└── outside.txt
```

reana-client:
```
==> Detected .gitignore file. Some files might get ignored.
==> SUCCESS: File /test/outside.txt was successfully uploaded.
==> SUCCESS: Symlink resolved to /test/outside.txt. Uploaded hard copy.
==> SUCCESS: File /test/dir/inside.txt was successfully uploaded.
```

reana-client-go:
```
==> SUCCESS: File test/dir/inside.txt was successfully uploaded.
==> ERROR: file 'test/dir_symlink' is a directory
```

audrium:
> Can you also reproduce this issue when dealing with symlinks?

I was able to reproduce it. In fact after spending some time trying to fix the issue I think it would be better to drop the support of symlinks because it could easily lead to troubles. It actually appears that Go's `filepath.Walk` function does not follow symlinks by design because it's difficult to get it correct avoiding infinite loops. 

As an experiment you could create a directory symlink which would point back to parent `test` folder and `reana-client` would break while uploading files! Also, while inspecting the current `reana-client` code I noticed it renames the symlinked files to their original names. It could also lead to troubles if a part of the workflow is designed to call a symlink by it's name (e.g while uploading `file_symlink.txt` it would be named `outside.txt` in the workflow).

As a quick solution to the problem I modified the code to detect the symlinked files and ignore it. We can discuss how we want to proceed with it and probably fix it in another PR as it appears to be a bit more difficult to do it right.

cmd/upload.go
```diff
@@ -0,0 +1,155 @@
...
+	for _, path := range args {
+		pathInfo, err := os.Stat(path)
+		if err != nil {
+			return nil, nil, err
+		}
+		if pathInfo.IsDir() {
+			dirs = append(dirs, path)
+		} else {
+			files = append(files, path)
+		}
```
mdonadoni, line 152: Symlinks cause some troubles also here, as they can be appended to `files`.

Given that `filepath.Walk` works also on files, I am wondering if we can avoid dividing paths into `files` and `directories`. This would make the code less complicated, as we wouldn't need the `filterDirectories` function. On the other hand, if a user specifies a file in `directories` (or a directory in `files`) in the specification we would need to check manually whether that is actually a directory or not to spot the error. What do you think?

mdonadoni: (I also don't know if you would prefer fixing this issues later in a new PR)

cmd/upload.go
```diff
@@ -0,0 +1,155 @@
...
+			if err != nil {
+				return err
+			}
+			// Check if a path is a symlink
+			if info.Mode()&os.ModeSymlink != 0 {
+				return nil
+			}
+			if !info.IsDir() {
+				files = append(files, path)
+			}
```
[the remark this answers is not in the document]
audrium: Nice suggestion, thanks!

cmd/upload.go
audrium: Good idea! I've changed the logic to use `filepath.Walk` on both files and directories and it works fine with less code, thanks! In this case even if user specifies a file in `directories` it works as expected.

>  On the other hand, if a user specifies a file in directories (or a directory in files) in the specification we would need to check manually whether that is actually a directory or not to spot the error. What do you think?

You mean we should report to a user if a file is found in `directories` and viceversa?

mdonadoni: > You mean we should report to a user if a file is found in `directories` and viceversa?

Yes, that is what I was referring to. I think we can discuss this in a separate issue, given that it also affects the python client!

mdonadoni (approved):

tiborsimko:
> As a quick solution to the problem I modified the code to detect the symlinked files and ignore it.

Yes, I don't think many people use the symlinks in sources, so as an MVP we could detect whether a file object is a symlink, and if it is, print a visible warning to the user that it'll be ignored (e.g. as we do for gitignore and friends) and continue with uploading the rest.

cmd/upload.go
tiborsimko: Yes, listing directories under `files` should be reported back to user, and vice versa. Since we have made that distinction, we should keep it.  

(Alternatively, we can respect both, so if a user uses only `files` and puts directories there, we can "silently" accept them. This is something that could pave the way towards using only one directive in the future.) 

BTW, regarding listing files and directories "unnecessarily", I have another observation. If a uses specifies the same file twice, it'll be uploaded twice, for example:

```
$ git diff
diff --git a/reana.yaml b/reana.yaml
index 4c69293..75d70e8 100644
--- a/reana.yaml
+++ b/reana.yaml
@@ -1,6 +1,11 @@
 version: 0.6.0
 inputs:
   files:
+    - code/gendata.C
+    - code/gendata.C
+    - code/gendata.C
+    - code/gendata.C
+    - code/gendata.C
     - code/gendata.C
     - code/fitdata.C
   parameters:
```

leads to:

```
$ reana-client run -w test
...
==> Uploading files...
==> Detected .gitignore file. Some files might get ignored.
==> SUCCESS: File /code/gendata.C was successfully uploaded.
==> SUCCESS: File /code/gendata.C was successfully uploaded.
==> SUCCESS: File /code/gendata.C was successfully uploaded.
==> SUCCESS: File /code/gendata.C was successfully uploaded.
==> SUCCESS: File /code/gendata.C was successfully uploaded.
==> SUCCESS: File /code/gendata.C was successfully uploaded.
==> SUCCESS: File /code/fitdata.C was successfully uploaded.
==> Starting workflow...
```

It would be nice to first make a list of all user-desired files and directories and walk through them, collect all the uploadable objects (avoiding symlinks), and then make a set out of them, so that we don't do any unnecessary upload the same object many times.

tiborsimko: P.S. The above upload behaviour was observed with REANA 0.8 clusters. Using the same client with REANA master branch clusters , one gets:

```
==> Starting workflow...
==> ERROR: Cannot start workflow test.1:
Input path declared multiple times: code/gendata.C
``` 

However, instead of throwing an error, it may be better to accept such a configuration and simply "deduplicate" the desired files and directories to be uploaded inside the client?

tiborsimko: (Especially because `reana-client validate` doesn't complain about the same file being listed multiple times. So, if we accept such a configuration as valid, we should allow processing it... Fully agree that these concerns are something for a new issue.)

audrium:
Updated PR with the following changes:
- detect whether a file path is a symlink and print a warning to the user 
- prohibit uploading the same file twice
- detect whether a directory is specified under files directive and vice versa and report to the user (with an error)

cmd/upload.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (o *uploadOptions) validateInputs(files, dirs []string) error {
+	for _, file := range files {
+		pathInfo, err := os.Stat(file)
+		if err != nil {
+			return err
+		}
+		if pathInfo.IsDir() {
+			return fmt.Errorf("directory is found in files directive: %s", file)
```
mdonadoni, line 155:
```suggestion
			return fmt.Errorf("found directory in `inputs.files`: %s", file)
```
WDYT?

cmd/upload.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+	for _, dir := range dirs {
+		pathInfo, err := os.Stat(dir)
+		if err != nil {
+			return err
+		}
+		if !pathInfo.IsDir() {
+			return fmt.Errorf("file is found in directories directive: %s", dir)
```
mdonadoni, line 164:
```suggestion
			return fmt.Errorf("found file in `inputs.directories`: %s", dir)
```
WDYT?

mdonadoni (approved):

PR #135, opened 2022-10-03
cli: implement download command

audrium:
closes https://github.com/reanahub/reana-client-go/issues/130

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -1,0 +1,190 @@
+/*
+This file is part of REANA.
+Copyright (C) 2022 CERN.
+
+REANA is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+package cmd
+
+import (
+	"archive/zip"
+	"bytes"
+	"fmt"
+	"io"
+	"path"
+	"reanahub/reana-client-go/pkg/config"
+	"reanahub/reana-client-go/pkg/displayer"
+	"reanahub/reana-client-go/pkg/fileutils"
+	"reanahub/reana-client-go/pkg/workflows"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const downloadDesc = `
+Download workspace files.
+
+The ` + "``download``" + ` command allows to download workspace files and directories.
+By default, the files specified in the workflow specification as outputs are
+downloaded. You can also specify the individual files you would like to
+download, see examples below.
+
+Examples:
+
+  $ reana-client download # download all output files
+
+  $ reana-client download mydata.tmp outputs/myplot.png
+
+  $ reana-client download -o - data.txt # write data.txt to stdout
+`
+
+const outputPathFlagDesc = `Path to the directory where files will be downloaded.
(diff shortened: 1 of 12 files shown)


cmd/download.go
```diff
@@ -0,0 +1,185 @@
...
+	return nil
+}
+
+// displayFileContent writes file(s) content to the standard output.
+func (o *downloadOptions) displayFileContent(
+	cmd *cobra.Command,
+	fileName string,
+	fileBuf *bytes.Buffer,
+) error {
+	if strings.Contains(fileName, ".zip") {
```
mdonadoni, line 130: There is a subtle difference between reana-client-go and the python client. In particular, when downloading a zip file with `-o -`, the zip is decompressed with reana-client-go, but not with reana-client.

```console
$ reana-client download -w workflow out/test.zip -o -
[... binary data ...]                                                             
$ ./reana-client-go download -w workflow out/test.zip -o -
[... contents of the zip file ...]
``` 
We are using the `Content-Type` header in reana-client to decide whether we should decompress the file or not (see [here](https://github.com/reanahub/reana-client/blob/e7175559735eb226662c47f67ec72993d16f573d/reana_client/api/client.py#L426-L429)), because that header is set to `application/zip` only when downloading multiple files at the same time, and not when downloading a single zip file.

What do you think about this?

audrium: In one sense you are right, current approach to always open and traverse the zip could lead to some troubles in case a file is big, but in the other hand if user specifies to display it, we should try to achieve that.
Therefore I have a mixed feelings about this, current logic in `reana-client` is to unzip multiple files, but to display zipped binary for a single file. But why we are not displaying a single file content? I think we should keep the current approach, since the zip could contain some readable information

cmd/download_test.go
```diff
@@ -0,0 +1,116 @@
...
+			params.cmd = "download"
+			testCmdRun(t, params)
+		})
+		t.Cleanup(func() {
+			// Remove all the temp files created by the test
+			err := os.RemoveAll(dirName)
+			if err != nil {
+				log.Fatal(err)
+			}
+		})
```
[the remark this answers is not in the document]
audrium: great catch, thanks

cmd/download.go
mdonadoni: The original idea of this flag was to have the same behaviour as `cat` (see https://github.com/reanahub/reana-client/issues/509, even though it doesn't talk about zip files), which just prints out the content of the file, even if it is just binary data. In general, using `-o -` with binary files does not make too much sense, unless the user wants to pipe the output to some other program.

I am (weakly) in favor of keeping the behavior of `reana-client`, that is to just print the binary data of the zip file. One of the reasons is that if we decide to decompress zip files, then we should also consider other file formats (e.g. `.rar`, `.tar.gz`, etc.).

@tiborsimko what is your opinion on this?

tiborsimko: Yes, I agree keeping the behaviour of `reana-client`, i.e. simply printing content regardless of its nature, to keep the analogy with `cat`.

audrium: thank you both for your input, I've modified the code accordingly

mdonadoni (approved):
Works well, Nice job!

PR #141, opened 2023-03-08
cli: fix typo in du command description and update license

giuseppe-steduto:


tiborsimko (commented):
Thanks, please just use "cli: fix typo in du command help" in the commit message.

(The license update is kind of self-understood, we have to do it with every commit...)

PR #144, opened 2023-06-06
interactive session: add message about autoclosure of the interactive session

giuseppe-steduto:
- Changes the `open` command to notify the user about the possible automatic closure of the interactive session due to inactivity
- Changes the `info` command to also display the maximum amount of days of inactivity before an interactive session gets automatically closed

Closes reanahub/reana-client#657


audrium (approved):
Looks nice! Small suggestion for the future: would be better to separate the swagger changes in a separate commit

cmd/open.go
```diff
@@ -114,5 +115,22 @@ func (o *openOptions) run(cmd *cobra.Command) error {
...
 	cmd.Println("It could take several minutes to start the interactive session.")
+
+	infoParams := operations.NewInfoParams()
+	infoParams.SetAccessToken(o.token)
+	infoResp, err := api.Operations.Info(infoParams)
+	if err != nil {
+		return nil
+	}
+	maxInactivityDays := infoResp.Payload.MaximumInteractiveSessionInactivityPeriod
+	if maxInactivityDays != nil {
```
mdonadoni, line 126:
```suggestion
	if maxInactivityDays != nil  && maxInactivityDays.Value != nil {
```
This will fail with `panic: runtime error: invalid memory address or nil pointer dereference` if `maximum_interactive_session_inactivity_period` is in the server's reply but `value` is `null`. Could you please add a testcase where this happens?

mdonadoni (commented):
Just left one small comment. Please also note that the copyright notice of the Go files needs to be updated!

PR #151, opened 2023-11-14
cli: implement share-status command

DaanRosendal:
Adds a new command to the CLI to retrieve whom a workflow is shared with.

Closes reanahub/reana-client#686


mdonadoni:
Included in https://github.com/reanahub/reana-client-go/pull/153

PR #152, opened 2023-11-20
cli: add include-last-command flag to list and status commands

giuseppe-steduto:
Add `--include-last-command` flag to the `list` and `status` commands
that, when set, will display info about the command currently being
executed by the workflow (or the last submitted command). In case there
is no info about the command, the step name will be displayed, if
possible.

Closes reanahub/reana-workflow-controller#486.


mdonadoni (approved):

mdonadoni:
No need to change API after performance improvements of https://github.com/reanahub/reana-db/pull/213

PR #156, opened 2024-01-29
fix(status): correctly display duration of stopped workflows (#156)

giuseppe-steduto:
Closes reanahub/reana-client#699


tiborsimko (approved):
Works nicely :+1: 

Before change:

```
DURATION
14608313
```

After change:

```
DURATION 
4057251
```

PR #164, opened 2024-11-11
feat(info): add missing Dask, engine and session information

### LINKED ISSUE #163: info: output Dask-related information
**Current behaviour**

After Dask workflow support has been added to REANA, the Python client `info` command replies with some Dask-related information to the user, whilst the Go client does not:

```console
$ rc info | grep -i dask
The number of Dask workers created by default: 2
The amount of memory used by default by a single Dask worker: 2Gi
The maximum memory limit for Dask clusters created by users: 16Gi
The maximum amount of memory that users can ask for the single Dask worker: 8Gi
Dask workflows allowed in the cluster: True

$ rcg info | grep -i dask
```

**Expected behaviour**

REANA Go client should also output the Dask-related information, on parity with what the Python client does.

Alputer:
Closes #163


cmd/info.go
```diff
@@ -99,6 +99,56 @@ func (o *infoOptions) run(cmd *cobra.Command) error {
...
+			displayInfoStringItem(cmd, p.DaskEnabled.Title, &p.DaskEnabled.Value)
+		}
+		if p.DaskEnabled != nil && strings.ToLower(p.DaskEnabled.Value) == "true" {
+			if p.DaskAutoscalerEnabled != nil {
+				displayInfoStringItem(cmd, p.DaskAutoscalerEnabled.Title, &p.DaskAutoscalerEnabled.Value)
+			}
+			if p.DaskClusterDefaultNumberOfWorkers != nil {
+				displayInfoStringItem(cmd, p.DaskClusterDefaultNumberOfWorkers.Title, &p.DaskClusterDefaultNumberOfWorkers.Value)
+			}
+			if p.DaskClusterMaxMemoryLimit != nil {
```
tiborsimko, line 139: You may want to amend the tests in order to have a case coverage for the new `info` endpoint replies?  Right new we have:

- testdata/inputs/info_big.json
- testdata/inputs/info_empty_inactivity_period.json
- testdata/inputs/info_small.json

You could introduce the workfow engine ones into "big" example, and you could introduce a new `info_dask.json` example for Dask variables, since the new Dask feature is optional...

tiborsimko: **question:** ‏The textual output of the Python client and the Go client is the same, however the order of outputted information differs, for example:

```console
$ rc info | tail -1
Yadage engine version: 0.20.1

$ rcg info | tail -1
The maximum number of workers that users can ask for the single Dask cluster: 20
```

Do we want to harmonise this at this time?

tiborsimko: **question:** ‏The JSON output of the Python client and the Go client differ with respect to null values: 

```console
$ rc info --json | jq > /tmp/z_py_j
$ rcg info --json | jq > /tmp/z_go_j
$ diff -u /tmp/z_py_j /tmp/z_go_j
--- /tmp/z_py_j 2025-01-21 13:46:00.209607732 +0100
+++ /tmp/z_go_j 2025-01-21 13:46:04.665618517 +0100
@@ -64,20 +64,17 @@
     "value": "False"
   },
   "kubernetes_max_memory_limit": {
-    "title": "Maximum allowed memory limit for Kubernetes jobs",
-    "value": null
+    "title": "Maximum allowed memory limit for Kubernetes jobs"
   },
   "maximum_interactive_session_inactivity_period": {
-    "title": "Maximum inactivity period in days before automatic closure of interactive sessions",
-    "value": null
+    "title": "Maximum inactivity period in days before automatic closure of interactive sessions"
   },
   "maximum_kubernetes_jobs_timeout": {
     "title": "Maximum timeout for Kubernetes jobs",
     "value": "1209600"
   },
   "maximum_workspace_retention_period": {
-    "title": "Maximum retention period in days for workspace files",
-    "value": null
+    "title": "Maximum retention period in days for workspace files"
   },
   "snakemake_engine_version": {
     "title": "Snakemake engine version",
```

Do we want to address this (admittedly minor) difference?

tiborsimko: **todo:** ‏For the commit log message, I would simply use 

```
feat(info): print information about Dask and workflow engines  (#164)
```

Motivation: the "missing" is from the point of view of compatibility with the Python client, but it's just a matter of time that that one was implemented first. It could have been the other way round... So from the point of view of the release notes of the go client, this item was not really "missing" before, because we did not have any Dask at the time yet. Hence it's nicer to just say that this was added, but not that it was "missing".

tiborsimko (approved):

PR #165, opened 2025-02-18
feat(info): display information about Dask worker threads

Alputer:
Following the recent change that makes the number of threads configurable for Dask workers, this commit adds the updates the `/api/info/` endpoint to include the two introduces parameters:

* `dask_cluster_default_single_worker_threads`: Defines the default number of threads per single Dask worker.
* `dask_cluster_max_single_worker_threads`: Sets an upper limit on the number of threads per single Dask worker.

Closes reanahub/reana#874

### DIFF
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -1,6 +1,6 @@
 /*
 This file is part of REANA.
-Copyright (C) 2022, 2024 CERN.
+Copyright (C) 2022, 2024, 2025 CERN.
 
 REANA is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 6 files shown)


cmd/info.go
```diff
@@ -1,6 +1,6 @@
 /*
 This file is part of REANA.
-Copyright (C) 2022, 2024 CERN.
+Copyright (C) 2022, 2024, 2025 CERN.
```
tiborsimko, line 3: **todo:** ‏Since this change principally affects the info command, the release news could be having that scope:

```text
feat(info): display information about Dask worker threads 
```

PR #169, opened 2025-12-18
 feat(etc): add bash and zsh shell completion files

tiborsimko:
Add shell completion scripts for `reana-client-go` generated using
Cobra's shell completion feature.

Add CI job to verify that committed shell completion files are
up-to-date with the current CLI commands.


michaelbuchar (approved):
LGTM!

PR #178, opened 2026-05-04
feat(logs): show Dask service logs

### LINKED ISSUE #768: feat(logs): show Dask service logs
### Current behaviour

For Dask workflows, the service logs are well exposed on the web interface:

Image

But not in Python or Go client:

```console
$ rc logs -w dask-coffea-serial-kubernetes
...
==> Job logs
==> Step: process
==> Workflow ID: 979a25e3-460b-4a29-bd92-680d08e566e6
==> Compute backend: Kubernetes
==> Job ID: reana-run-job-22af4bc4-4d0f-4cc9-a74f-219e6f8cf1d9
==> Docker image: docker.io/coffeateam/coffea-dask-cc7:0.7.22-py3.10-g7f049
==> Command: python analysis.py
==> Status: finished
==> Started: 2026-05-04T08:57:58
==> Finished: 2026-05-04T08:58:53
==> Logs:
job: :
 Matplotlib created a temporary cache directory at /tmp/matplotlib-8z22hpf3 because the default path (/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.
all events 53446198
number of chunks 534


Completed
```

### Expected behaviour

The CLI clients should show the service logs just as the Web UI does.

tiborsimko:
The REST API already exposes Dask scheduler and worker pod logs under the `service_logs` key of the workflow logs response, and the Web UI surfaces them in a dedicated tab. The CLI silently dropped the field, leaving Dask users without a terminal-side view of their cluster logs.

Render the captured pod logs in `reana-client logs` after the workflow engine logs and before the job logs.

Closes reanahub/reana-client#768


tiborsimko:
Note: the CI test fail due to bad logs testing logic, which is fixed in #179. So that one should go in for the review first...

cmd/logs.go
```diff
@@ -449,6 +457,57 @@ func displayHumanFriendlyLogs(cmd *cobra.Command, logs logs, steps []string) {
 		cmd.Println(*logs.EngineSpecific)
 	}
 
+	if len(logs.ServiceLogs) > 0 {
```
michaelbuchar: ### AI Suggestion - similar as https://github.com/reanahub/reana-client/pull/769#discussion_r3188450966

The code checks only if len(logs.ServiceLogs) > 0 before printing the Service logs section. The server can legitimately return a payload where a Dask service exists but has no stored entries yet, e.g. {"service_logs": {"dask-service-123": []}}. In that case the CLI will print Service logs and Service: dask-service-123, then nothing else. It should either skip empty entries or print an explicit message like Service dask-service-123 emitted no logs.

tiborsimko: Thanks, addressed. The human-friendly renderer now builds the list of printable services from entries with at least one service-log item, so a payload such as `{"service_logs": {"dask-service-123": []}}` no longer prints an empty `Service logs` section.

I also added a regression fixture for this case. Entries with an existing component but empty `content` still keep the explicit `Component ... emitted no logs.` message.

michaelbuchar (approved):
👍

PR #179, opened 2026-05-04
fix(logs): match followed step by job_name

tiborsimko:
When `--follow --filter step=X` is used, getLogsWithStatus() delegated to getFirstJob(), which simply returned an entry from the response map without consulting `step` at all. With Go's randomised map iteration, even the chosen "first" job varied between processes; making the choice deterministic by sorting keys (an earlier attempt) only papered over that.

Fix the helper to filter by jobLogItem.JobName matching the requested step, returning nil (-> "step X not found" error) when no entry matches. When several jobs legitimately share the same step name (some workflow engines may produce many job rows under one job_name for scatter or parallel steps), tie-break on the lowest map key so the choice is stable across calls.

This also fixes the "follow_job_with_multiple_steps" etc CI flake. The test fixture carries job1 (status finished) and job2 (status running); the test passes step=job1, and findJobByStep() now correctly returns job1 every time, so the follow loop sees status=finished and exits. The flake had been hidden by Go's per-process map seed staying stable for tiny maps within a single test run.

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -17,6 +17,7 @@ import (
 	"reanahub/reana-client-go/pkg/config"
 	"reanahub/reana-client-go/pkg/displayer"
 	"reanahub/reana-client-go/pkg/filterer"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


michaelbuchar (approved):
LGTM!

PR #180, opened 2026-05-13
feat(quota): expose quota period in CLI commands

### LINKED ISSUE #771: cli: expose quota period and replenishment window
### Summary

Expose periodic quota information in the CLI so that users can understand which accounting window their current usage applies to and when their quota is expected to replenish.

At the moment, the Web UI shows the current CPU quota period window, but the CLI does not surface comparable information.

### Problem

For deployments using periodic CPU quota accounting, the current CLI output does not tell the user:

- whether a periodic quota policy is active
- which quota period the shown CPU usage applies to
- when the current accounting window ends / when quota is expected to replenish

Examples:

```console
$ reana-client quota-show --resource cpu -h
30s out of 16m 40s used (3%)
```

The output above does not indicate the active quota period.

Similarly, `reana-client info` does not expose server-defined quota period information.

### Expected behaviour

Achieve CLI feature parity with the Web UI for quota-period visibility.

Possible improvements:

- `reana-client info` should show whether a periodic CPU quota policy is active
- `reana-client quota-show --resource cpu -h` should display the active accounting window
- the shown period should make it clear when users can expect quota replenishment

For example, the CLI could show something like:

```console
$ reana-client quota-show --resource cpu -h
30s out of 16m 40s used (3% in the period from 2026-05-01 to 2026-08-31) 
```

### Notes

- The backend already exposes `quota_period_months` and `quota_period_start_at`
- The Web UI already renders this information
- A matching follow-up should likely be done for `reana-client-go` as well
- This is intentionally split from the main quota-period backend/UI PR set to keep review scope manageable

michaelbuchar:
Closes reanahub/reana-client#771


cmd/quota_show.go
```diff
@@ -171,6 +175,7 @@ func (o *quotaShowOptions) run(cmd *cobra.Command) error {
 		displayQuotaResourceUsage(
 			resource.Health,
 			resource.Stats["usage"], resource.Stats["limit"],
+			formatQuotaPeriodWindow(resource),
 			o.humanReadable, cmd.OutOrStdout(),
 		)
 	} else if !reportExists || report.Raw <= 0 {
```
tiborsimko, line 188: 🤖 **Show CPU period suffix when `-h` is used with `--report usage|limit` (LOW)**

The default mode (`quota-show -r cpu`) now displays the active CPU accounting window, but the explicit `--report` paths drop it:

```
$ reana-client-go quota-show --resource cpu
30072 out of 6789000 used (0% in the period from 2026-05-15 to 2026-07-15)

$ reana-client-go quota-show --resource cpu --report usage -h
30s
```

A user reading `30s` in isolation can reasonably mis-interpret it as lifetime usage. The natural rule is *raw output is scriptable, `-h` is for humans* — so the period belongs in `-h` output for both `--report usage` and `--report limit` on CPU (the limit applies to one period, so the window is part of what the limit means).

Proposed behaviour for the `--resource cpu` path on a user with periodic CPU accounting:

| Flags | Today | Proposed |
|---|---|---|
| `--report usage` | `30000` (raw) | unchanged |
| `--report usage -h` | `30s` | `30s in the period from 2026-05-15 to 2026-07-15` |
| `--report limit` | `6789000` (raw) | unchanged |
| `--report limit -h` | `1885h 50m` | `1885h 50m in the period from 2026-05-15 to 2026-07-15` |

For disk and for CPU users without periodic accounting, the suffix should not appear (today's behaviour). Raw output stays untouched in all cases so scripts keep working.

Suggested fix — extend the `else if o.humanReadable:` branch (line 183 in this file, just below this comment's anchor):

```go
} else if o.humanReadable {
    msg := report.HumanReadable
    if o.resource == "cpu" {
        if window := formatQuotaPeriodWindow(resource); window != "" {
            msg = fmt.Sprintf("%s in the period from %s", msg, window)
        }
    }
    cmd.Println(msg)
}
```

New test cases for `cmd/quota_show_test.go`:

- `--report usage -h` for a periodic CPU user → expect the suffix
- `--report usage -h` for a non-periodic CPU user → no suffix (regression guard)
- `--report usage` (no `-h`) for a periodic CPU user → still just the raw number (regression guard)
- `--report limit -h` for a periodic CPU user → expect the suffix
- `--report usage -h` for disk → no suffix

The same change should land in `reana-client` PR #772 (Python) for symmetry — equivalent comment posted there.

go.mod
```diff
@@ -1,6 +1,6 @@
 module reanahub/reana-client-go
 
-go 1.25.9
+go 1.25.10
```
tiborsimko, line 3: 🤖 **Go toolchain bump is unrelated to the quota-period feature (LOW)**

This PR bumps `go 1.25.9 → 1.25.10` here, plus the matching versions in `.mise.toml` and `.github/workflows/ci.yml`, plus the transitive `golang.org/x/{net,sync,sys,text}` updates in `go.sum`. None of these are needed for the periodic-quota display — they're routine maintenance that happens to have ridden along on the feature branch.

Mixing the two has two costs:

1. **Bisect-hostility.** If a regression turns up later in `golang.org/x/net 0.53.0` or in 1.25.10, `git bisect` lands on this commit and you have to read past 300+ lines of quota-display code to find the dep bump that actually caused it. Conversely, if `feat(quotas): display individual user quota period in quota-show` ever needs to be reverted, the revert also reverts the toolchain.
2. **Reviewer cognitive load.** A focused PR is easier to assess.

Suggestion — split into two PRs, e.g.:

- `build(deps): upgrade Go toolchain to 1.25.10 and transitive x/* deps` (the changes to `go.mod`, `go.sum`, `.mise.toml`, `.github/workflows/ci.yml`).
- `feat(quotas): display individual user quota period in quota-show` (everything else).

If rebasing now is more friction than it's worth, optional: keep the bundle but mention in the PR description that the dep bumps are unrelated, so a future bisecter has a breadcrumb.

Not a blocker — the bundle works; this is about future maintainability.

cmd/info_test.go
```diff
@@ -119,3 +147,55 @@ func TestInfo(t *testing.T) {
...
+	output, err := ExecuteCommand(rootCmd, "info", "-t", "1234")
+	if err != nil {
+		t.Fatalf("Got unexpected error '%v'", err)
+	}
+
+	lines := strings.Split(strings.TrimSpace(output), "\n")
+	if len(lines) == 0 {
+		t.Fatalf("Expected info output, got empty output")
+	}
+	lastLine := strings.TrimSpace(lines[len(lines)-1])
```
tiborsimko: 🤖 **Asserting on the literal last line couples the test to unrelated output (NIT)**

This test pins `"CPU quota period: Disabled"` as the very last printed line of `info`. The intent — *"the quota period info comes after the standard info block"* — is reasonable, but expressing it as `lines[len(lines)-1]` makes the test fail for any future change that simply adds *another* `info` field below the quota period section. Examples that would break this:

- A new engine version line appended after `YadageEngineVersion`.
- A new cluster-capability line.
- A `Maximum job nodes` style entry from a future PR.

None of those are regressions in the behaviour this test is trying to guard.

Two less brittle ways to express the same invariant:

1. **Assert relative order between two specific lines.** Find the position of `"List of supported compute backends:"` (or any other stable info-block line) and the position of `"CPU quota period:"`, and assert the second is later. Survives unrelated additions.

   ```go
   infoIdx := -1
   quotaIdx := -1
   for i, line := range lines {
       trimmed := strings.TrimSpace(line)
       if strings.HasPrefix(trimmed, "List of supported compute backends:") {
           infoIdx = i
       }
       if strings.HasPrefix(trimmed, "CPU quota period:") {
           quotaIdx = i
       }
   }
   if infoIdx < 0 || quotaIdx < 0 || quotaIdx <= infoIdx {
       t.Fatalf("Expected CPU quota period to be printed after the info block, got order info=%d quota=%d in:\n%s", infoIdx, quotaIdx, output)
   }
   ```

2. **Assert that the quota period section is the trailing one *among the lines we care about*.** Skip blank/footer lines and assert the last non-blank line that starts with a known info-field prefix is the quota period one. Slightly more code than option 1, but extends to a future "and `Current CPU quota period:` comes right after `CPU quota period:`" assertion.

Option 1 is enough to capture the intent and is the smaller change. If you want to keep the current assertion, please rename the test to make the *strict* expectation explicit (e.g. `TestInfoQuotaPeriodIsLiteralLastLine`) so a future maintainer who breaks it by appending a new field knows immediately whether their change is the regression or whether the test is over-tight.

cmd/quota_show.go
michaelbuchar: Fixed

cmd/info_test.go
michaelbuchar: Fixed

cmd/quota_period.go
```diff
@@ -0,0 +1,95 @@
...
+		PeriodLabel: "Disabled",
+	}
+	if cpuResource.QuotaPeriodMonths != nil &&
+		*cpuResource.QuotaPeriodMonths > 0 {
+		months := *cpuResource.QuotaPeriodMonths
+		unit := "months"
+		if months == 1 {
+			unit = "month"
+		}
+		info.PeriodLabel = fmt.Sprintf("Every %d %s", months, unit)
```
tiborsimko: 🤖 **`info` quota fields should be machine-parseable scalars, not English prose (LOW)**

Today's `info` output for the quota period is human-readable but not scriptable:

```
$ reana-client-go info | grep -i quota
CPU quota period: Every 2 months
Current CPU quota period: 2026-06-04 to 2026-09-04

$ reana-client-go info --json | jq -S '.cpu_quota_period'
{
  "title": "CPU quota period",
  "value": "Every 2 months"
}
```

A user wanting to alert when their period changes, confirm a deployment-wide rollout, or feed the value into another tool has to regex-parse `"Every N months"` / `"Disabled"` — and the `"Disabled"` branch is a different shape than the numeric branch, so the parser has to handle two cases. Likewise, `"2026-06-04 to 2026-09-04"` requires splitting on `" to "` to extract either endpoint.

Suggested fix — break the structured fields out of the prose and use scalar values, matching the convention `quota.default_cpu_limit: 0` already uses (`0 = disabled / unlimited`):

```
$ reana-client-go info | grep -i quota
CPU quota period in months: 2
Current CPU quota period start: 2026-06-04
Current CPU quota period end: 2026-09-04

$ reana-client-go info --json | jq -S '.cpu_quota_period_months.value, .current_cpu_quota_period_start.value, .current_cpu_quota_period_end.value'
2
"2026-06-04"
"2026-09-04"
```

For a user with no periodic CPU accounting, `cpu_quota_period_months.value` is `0`, and the start/end fields are omitted (or `null` in JSON if you prefer always-present keys). The plain-text rendering for that case becomes just `CPU quota period in months: 0`, with no start/end lines — uniform parse rule, no special case.

This maps cleanly to existing `info` precedent: each field has a `title` + a scalar `value`. The current `"Every 2 months"` is a derived display string, which doesn't belong in the structured payload.

For `buildCPUQuotaPeriodInfo` in this file, the shape would shift from a single `PeriodLabel`/`CurrentWindow` pair to scalar months + two date strings (start, end). The corresponding `displayQuotaPeriodInfo` and `buildInfoOutputMap` in `cmd/info.go` would emit three keys instead of two, and the plain-text printf lines change accordingly. If you address the period-end-date-duplication MEDIUM comment (from reana-client #772) by having the server return `quota_period_end_at`, the local `addMonths` call drops out and `cmd/quota_period.go` shrinks to just date formatting.

For `quota-show`, the existing `... in the period from X to Y` rendering already exposes both dates, so no change needed there — only `info` needs this.

Cross-reference: same fix applies to `reana-client` PR #772 in `reana_client/utils.py:build_cpu_quota_period_info` — equivalent comment posted there with a concrete Python diff.

go.mod
michaelbuchar: Fixed

cmd/quota_period.go
michaelbuchar: Fixed

cmd/quota_period.go
```diff
@@ -0,0 +1,105 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type quotaPeriodInfo struct {
+	PeriodMonths *int64
+	StartDate    string
+	EndDate      string
+}
+
+func addMonths(dt time.Time, months int) time.Time {
```
tiborsimko: 🤖 **Add the same cross-component sync NOTE that landed on `reana-client` (NIT)**

The Python counterpart in `reana_client/utils.py:_add_months` now carries a comment naming the other implementations that must stay in lockstep:

```python
# NOTE: Keep this month-boundary arithmetic in sync with reana_db/utils.py
# (_add_months), reana-client-go's addMonths helper, and reana-ui's quota
# period window calculation.
```

A future maintainer working only in this repo doesn't see that comment. A small symmetric note here would close the loop:

```go
// NOTE: Keep this month-boundary arithmetic in sync with
// reana_db/utils.py (_add_months), reana-client's _add_months helper,
// and reana-ui's quota period window calculation.
func addMonths(dt time.Time, months int) time.Time {
```

One-liner; no functional change. Just keeps the duplication visible from both sides.

michaelbuchar: Fixed

cmd/quota_period.go
```diff
@@ -0,0 +1,105 @@
+/*
```
tiborsimko: 🤖 **`cmd/quota_period.go` breaks the `cmd/foo.go = one CLI subcommand` convention (LOW)**

Every other file under `cmd/` corresponds 1:1 to a runnable subcommand (`cmd/info.go` → `reana-client-go info`, `cmd/quota_show.go` → `reana-client-go quota-show`, `cmd/secrets_add.go` → `reana-client-go secrets-add`, …). The only exception today is `cmd/root.go`, which is the root Cobra wiring.

This new file is different: there's no `reana-client-go quota-period` subcommand. It's a shared helper module — `quotaPeriodInfo` struct, `addMonths`, `formatQuotaPeriodDate`, `getQuotaPeriodDateRange`, `formatQuotaPeriodWindow`, `buildCPUQuotaPeriodInfo` — consumed by `cmd/quota_show.go` and `cmd/info.go`. Living under `cmd/` makes a future reader wonder which subcommand it implements.

Three reasonable ways to fix it, ordered by ascending invasiveness:

1. **Fold the helpers into `cmd/quota_show.go`.** Lowest friction — `quota_show.go` becomes the canonical home for all quota-show-related code, and the file grows by ~95 lines (still well-sized). There's already a precedent in this PR: `cmd/info.go` reaches into `cmd/quota_show.go` for `parseQuotaInfo`, so "info-side quota logic borrows from `quota_show.go`" is already the established pattern. This is the minimum change that fixes the convention break.

2. **Move into an existing `pkg/` subpackage.** `pkg/displayer/quota_period.go` is the closest semantic match — the helpers all feed display strings — and `pkg/formatter/` is another candidate. Useful if you anticipate more quota-related display helpers in the future.

3. **New `pkg/quotaperiod/` subpackage.** Cleanest separation. Exposes a public surface (`QuotaPeriodInfo`, `BuildCPUQuotaPeriodInfo`, `FormatQuotaPeriodWindow`, `GetQuotaPeriodDateRange`) and keeps `addMonths`/`formatQuotaPeriodDate` package-private. Worth it only if the feature grows.

Option 1 is probably the right call for now given the size and the existing `info.go ↔ quota_show.go` coupling. Any of the three is better than leaving the file under `cmd/`.

cmd/quota_show.go
```diff
@@ -41,7 +41,11 @@ Examples:
 `
 
 type quotaResource struct {
-	Health string `json:"health"`
+	Health             string             `json:"health"`
+	Limit              *quotaResourceStat `json:"limit"`
+	QuotaPeriodMonths  *int64             `json:"quota_period_months"`
+	QuotaPeriodStartAt *string            `json:"quota_period_start_at"`
```
tiborsimko, line 48: 🤖 **Tighten commit scope to match the project's `cmd/<scope>.go` convention (NIT)**

Looking at scopes used on master (`feat(info)`, `feat(logs)`, `refactor(info)`, `fix(logs)`, `fix(cmd)`, `feat(client)`, `feat(etc)`, `build(go)`, `chore(master)`), the convention is **singular subcommand name** (matching the `cmd/<scope>.go` filename) or **package area**. There's no precedent for a plural feature group like `quotas`.

For this PR's three commits:

| Current | Suggested |
|---|---|
| `feat(quotas): show quota period in quota-show (#180)` | `feat(quota-show): show quota period (#180)` |
| `feat(info): display server-side quota period settings (#180)` | unchanged ✓ |
| `build(go): upgrade to Go 1.25.10 (#180)` | unchanged ✓ |

The rename also lets you drop the redundant "in quota-show" from the description, so the headline reads cleaner at 47 chars.

Worth amending before merge since commitlint checks the scope and the message is permanent once it lands on master.

cmd/quota_period.go
michaelbuchar: Fixed with option 1

cmd/quota_show.go
michaelbuchar: Fixed the commit message

tiborsimko (approved):

PR #181, opened 2026-05-19
feat(info): display minimum Kubernetes job user ID

### LINKED ISSUE #951: helm: allow cluster admins to configure the minimum accepted `kubernetes_uid` value
## Current behaviour

User-supplied `kubernetes_uid` workflow hint values are accepted as long as they are `>= 100`. This is hard-coded in REANA ecosystem, e.g. in the `reana-job-controller` component. A REANA deployment may want to strengthen the policy to allow only higher user IDs, similar to OpenShift.

  ## Expected behaviour

The minimum accepted `kubernetes_uid` value should be configurable per REANA deployment via the Helm value files, for example by means of an environment variable such as  `REANA_KUBERNETES_JOBS_MIN_USER_UID`.

tiborsimko:
Render the new `kubernetes_min_user_uid` field from the `/info` endpoint so that users can verify the cluster-configured minimum runtime container user ID accepted in `kubernetes_uid` workflow hints.

Closes reanahub/reana#951

### DIFF
--- a/client/operations/info_responses.go
+++ b/client/operations/info_responses.go
@@ -302,6 +302,9 @@ type InfoOKBody struct {
 	// kubernetes max memory request
 	KubernetesMaxMemoryRequest *InfoOKBodyKubernetesMaxMemoryRequest `json:"kubernetes_max_memory_request,omitempty"`
 
+	// kubernetes min user uid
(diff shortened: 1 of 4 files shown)


michaelbuchar (approved):
LGTM!

PR #182, opened 2026-06-05
feat(info): display workflow image vetting settings

### LINKED ISSUE #728: validate: check workflow specification for unathorised container images
In some deployment scenarios we may want to allow running only those workflows which use vetted authorised container images.

(This is similar to how we allow to run only certain vetted Jupyter notebook images by users, see `interactive_sessions.environments.jupyter.recommended` Helm value.)

To address these deployment scenarios, it would be useful to:

- allow REANA cluster administrators to specify which images can be used by users in their runtime workflows;
- enrich `reana-client info` to report to users which images are authorised;
- amend `reana-client validate` to report any troubles early in case a user tries to use an unauthorised image;
- make strong check on the REANA server side not to accept user workflows if the workflow uses an unauthorised image for some of its steps.

The implementation could be similar to how we handle Jupyter images, i.e. via a Helm value. However, the list of authorised images may grow, so we may have to be ready to handle potentially large number of allowed items in the vetted container image list.

CameronMcClymont:
Closes reanahub/reana-server#728


cmd/info.go
```diff
@@ -202,6 +202,16 @@ func (o *infoOptions) run(cmd *cobra.Command) error {
...
 		}
+		if p.VettedContainerImagesAllowlist != nil {
+			displayInfoSliceItem(cmd, p.VettedContainerImagesAllowlist.Title, p.VettedContainerImagesAllowlist.Value)
+		}
+		if p.VettedContainerImagesEnabled != nil {
+			value := "False"
+			if p.VettedContainerImagesEnabled.Value {
+				value = "True"
+			}
+			cmd.Printf("%s: %s\n", p.VettedContainerImagesEnabled.Title, value)
```
tiborsimko, line 213: **🤖 Could add a false-valued fixture for `vetted_container_images_enabled` (NIT)**

The current info fixture covers the `True` output. Since this is the first boolean-valued `/info` field handled by a one-off formatter, it would be useful to also assert the `False` rendering, e.g. in the missing-fields/small fixture or a tiny dedicated case.

Suggested fix — extend `testdata/inputs/info_small.json` with `"vetted_container_images_enabled": {"value": false, "title": "Whether container image vetting is enabled"}` and add `"Whether container image vetting is enabled: False"` to the `missing fields` expectations in `cmd/info_test.go`.

tiborsimko, line 213: **🤖 `/info` now mixes real booleans with string-encoded booleans (NIT, server-side)**

`vetted_container_images_enabled.value` is a JSON boolean, while older toggles such as `dask_enabled` and `interactive_sessions_custom_image_allowed` remain `"True"`/`"False"` strings. The Go client handles this correctly here (hence the manual `True`/`False` conversion above), but it may be worth deciding on one convention for future `/info` fields.

Suggested fix — raise on reanahub/reana-server#739 (e.g. converge older toggles on `BooleanInfoValue`, or document the rationale). No change needed in this Go client PR.

tiborsimko:
BTW this is something for the REANA cluster (commons): the current output looks like:

```console
$ rcg info
...
List of container images allowed in user workflows: docker.io/library/python:2.7-slim, docker.io/library/python:2.8-slim
Whether container image vetting is enabled: True
```

The word "vetted" is used only once, even though the property is named that way, and also "Whether container..." reads a bit different from the rest such as Dask.

What about printing:

```console
$ rcg info
...
Vetted container images required for user workflows: True
List of vetted container images allowed in user workflows: docker.io/library/python:2.7-slim, docker.io/library/python:2.8-slim
```

tiborsimko (approved):

PR #184, opened 2026-06-26
feat(open): scope interactive session secrets

### LINKED ISSUE #978: Allow restricting which user secrets are exposed to workflow jobs and interactive sessions
After #805, user secrets can be propagated to interactive sessions. Today, secrets are generally exposed broadly to workflow job pods, Dask workers, and interactive sessions. This is convenient, but it also means pods receive secrets they do not actually need.

## Problem

There is currently no least-privilege way for a workflow author or an interactive-session caller to restrict which user secrets are exposed to a given pod/session.

We should keep the current default behaviour for backward compatibility, but provide an explicit opt-in mechanism to reduce the exposed secret set when desired.

## Proposed solution

Introduce an optional `secret_names` allowlist.

Behaviour:

- If `secret_names` is omitted: keep current behaviour and expose all user secrets.
- If `secret_names` is an empty list: expose no user secrets.
- If `secret_names` is provided: expose only the named secrets.
- If a requested secret name does not exist: reject the request with a clear validation error.

Implementation approach:

- Filter by secret name once in `UserSecrets`, then derive env refs and mounted file items from the filtered subset.
- Preserve the current behaviour for Kerberos/VOMS/Rucio-enabled paths for now, i.e. do not scope secrets there yet.

## User-facing scope

Interactive sessions:

- `reana-client open ... --secret-name NAME` (repeatable)
- `reana-client open ... --no-secrets`
- `reana-client-go open ... --secret-name NAME` (repeatable)
- `reana-client-go open ... --no-secrets`

Workflow authoring:

- Serial: step-level `secret_names: ["alpha", "beta"]` in `reana.yaml`
- Dask: `workflow.resources.dask.secret_names: ["alpha", "beta"]` in `reana.yaml`
- Snakemake: rule-level `resources: secret_names="alpha,beta"`
- Yadage: step `environment.resources` entry such as `- secret_names: ["alpha", "beta"]`

## Out of scope for this issue

- CWL support
- Secret scoping for Kerberos/VOMS/Rucio-enabled jobs

These can be addressed in follow-up work once the base scoping model is agreed.

## Acceptance criteria

- Existing workflows and interactive-session flows remain unchanged when `secret_names` is not provided.
- Interactive sessions can request either an explicit secret allowlist or no secrets at all.
- Serial, Snakemake, Yadage, and Dask authors can specify the secret subset for their jobs/workers.
- Empty allowlists result in no secrets being exposed.
- Unknown secret names produce a clear user-facing error.
- Kerberos/VOMS/Rucio-enabled paths preserve current behaviour.

## Affected components

- `reana-commons`
- `reana-job-controller`
- `reana-workflow-controller`
- `reana-client`
- `reana-client-go`
- `reana-workflow-engine-serial`
- `reana-workflow-engine-snakemake`
- `reana-workflow-engine-yadage`
- `reana-server`

michaelbuchar:
Add secret scoping flags and request handling for interactive sessions, including explicit no-secrets support and omitted-field behaviour that preserves server-side inheritance semantics.

Closes reanahub/reana#978


etc/bash_completion.d/reana-client-go
```diff
@@ -11,7 +11,7 @@
 #     . /path/to/reana-client-go/etc/bash_completion.d/reana-client-go
 #
 #   Option 2: Copy to system completion directory:
-#     cp /path/to/reana-client-go/etc/bash_completion.d/reana-client-go \
+#     cp /path/to/reana-client-go/etc/bash_completion.d/reana-client-go \\
```
tiborsimko, line 14: **🤖 PR184-01: Regenerated completion files corrupt an installation example (NIT)**

The regenerated Bash and Zsh completion files change the example `cp` command's line continuation from a single backslash to two literal backslashes. Copying the documented two-line command therefore passes a literal backslash argument instead of continuing the command, so the installation example no longer works as written.

Suggested fix — restore the single backslash in both completion-file headers and ensure the generation/update process preserves these manually maintained introductory comments.

cmd/open.go
```diff
@@ -90,6 +102,18 @@ func newOpenCmd() *cobra.Command {
 		"Name or UUID of the workflow. Overrides value of REANA_WORKON environment variable.",
 	)
 	f.StringVarP(&o.image, "image", "i", "", openImageFlagDesc)
+	f.StringSliceVar(
```
tiborsimko, line 105: **🤖 PR184-02: `--secret-name` has different comma semantics in the two clients (NIT)**

`StringSliceVar` treats `--secret-name a,b` as two values, while the Python client and the documented repeatable `NAME` contract treat each option occurrence as one literal secret name. The clients therefore disagree for a valid secret whose name contains a comma, and the Go command accepts an additional comma-separated syntax that its help text does not describe.

Suggested fix — use Cobra's `StringArrayVar` so each option occurrence contributes exactly one name, matching the Python client. Add coverage for repeated values and a comma-containing value.

PR #185, opened 2026-06-29
feat(validation): upload scoped workflow snapshots in Go client

### LINKED ISSUE #8: Move workflow specification building and validation to the server
## Background

Today, validating a workflow specification requires the client to call the corresponding workflow engine (CWL, Serial, Snakemake, Yadage) to build the specification, then validate it locally. The server receives this specification and executes it as-is.

## Problem

This client-side approach has two consequences:

- Client language lock-in. Supporting a new client requires reimplementing every supported workflow engine in that language. This is why reana-client-go currently cannot create or submit workflows — porting all engines to Go is impractical.
- No security boundary. The server trusts a specification generated on the client without re-checking it. Client-side validation is a quality-of-life feature that catches mistakes early, not a guarantee the server can rely on.

## Proposal

Move specification building and validation to the server, so any client only needs to send a workflow snapshot (source files) and let the server do the rest.

Because building a specification means evaluating the workflow file, which can execute arbitrary code (e.g. Snakemake run: blocks, Jinja templating), this must happen inside a sandbox with least-privilege access — otherwise the server itself becomes remotely exploitable by a malicious specification.

## Acceptance Criteria

- [ ] Specification building and validation happen server-side, so a client only needs to submit workflow files, not run the corresponding engine
- [ ] Specification building and validation run in a sandboxed, least-privilege environment isolated from the rest of the server
- [ ] The sandbox has network access to fetch external dependencies referenced by the workflow (e.g. git+https CWL tools)

Sohn123:
Add create and validate commands using the generated Swagger client and bounded, streamed workflow snapshots. The Go client mirrors the Python validation-scope and path-security contracts, keeps image pulls local, uses bounded control responses and explicit operation deadlines, and uploads restart replacements canonically. Input datasets are excluded from validation snapshots. Restart replaces only `reana.yaml` and reuses workflow sources already present in the workspace.

Closes reanahub/reana-workflow-validator#8


pkg/specbundle/specbundle.go
```diff
@@ -0,0 +1,230 @@
...
+
+// addWorkflowFileMembers adds workflow.file to the bundle, together with its
+// whole sub-tree when it lives in a sub-directory (to capture co-located
+// includes/imports/stage files). A top-level workflow file is added on its own
+// rather than gathering the entire specification directory.
+func addWorkflowFileMembers(
+	members map[string]string,
+	baseDir, workflowFile string,
+) error {
+	abs := filepath.Join(baseDir, workflowFile)
```
tiborsimko: **🤖 SNDBX-04: Client spec-bundle assembly has no path-escape or symlink containment (MEDIUM)**

*Component:* reana-client — `reana_client/api/client.py:249-258,261-291,369-374`;
reana-client-go — `pkg/specbundle/specbundle.go:112-127,137-168,180-200`

Both clients build the upload bundle from `workflow.file` (and
`inputs.parameters.input`) in the user's `reana.yaml` without validating that the
referenced paths stay inside the spec directory, and without skipping symlinks.

In `_add_workflow_file_members` (`reana_client/api/client.py:261-291`), a
`workflow.file` that escapes the base dir does not just read one file — it walks
the **parent tree**. For `workflow.file: ../secret.txt`, `workflow_dir` becomes
the parent of `base_dir`, the `same_dir or not isdir` guard is False, and control
falls into `os.walk(workflow_dir)`, bundling the whole parent directory.
`workflow.file: /etc/passwd` walks `/etc`; `workflow.file: /` walks the entire
filesystem. `_add_file_member` (`client.py:249-258`) gates on `os.path.isfile`,
which **follows symlinks**, so symlink targets are read and uploaded too. The
files are opened and sent at `client.py:369-374`. The Go client is an exact port
with the same behaviour: `os.Stat` follows symlinks (`specbundle.go:114`),
`filepath.Rel` is computed with no `..` check (`:118`), and
`filepath.Walk(workflowDir)` runs on the parent on escape (`:146`).

**Severity / threat model.** Under the normal model — a client talking to its own
trusted REANA server — this is MEDIUM, not HIGH. The server rejects any
`..`/absolute member name at staging *and* `rmtree`s the staging dir in `finally`
(`reana_server/rest/workflows.py:113-144`), and an escaping path necessarily
produces a `..`-prefixed member name, so escaping members are never persisted to
a workspace nor returned to the client — the "malicious spec author receives the
victim's file" path does not complete. What remains is still block-worthy: the
client **reads and transmits** out-of-project files (including symlink targets)
over the wire before rejection (a privacy/over-share leak, and a real exposure if
the server logs member content or an operator can read the transient upload
spool); `validate`, which users reasonably expect to be a *read-only local lint*,
silently uploads; and a crafted `workflow.file` can make the client walk an
enormous tree (parent dir, `/`) — a client-side hang/DoS footgun. It rises to HIGH
only when the client is pointed at an untrusted/multi-tenant server.

Suggested fix — validate bundle member paths **client-side before opening**, in
both clients: relative only, normalized, no `..` segment, not absolute, and no
symlink escape from the spec base directory (reuse `is_regular_path` in the Python
client, mirror it in Go). Consider a per-file/total size cap and an explicit error
on rejected members so the user understands why.

pkg/specbundle/specbundle.go
```diff
@@ -0,0 +1,230 @@
...
+	workflowDir := filepath.Dir(abs)
+
+	info, err := os.Stat(workflowDir)
+	if filepath.Clean(workflowDir) == baseDir || err != nil || !info.IsDir() {
```
tiborsimko: **🤖 SNDBX-06: Top-level workflow file with relative includes is under-bundled — common layouts fail validation (MEDIUM)**

*Component:* reana-client — `reana_client/api/client.py:261-291`,
`tests/test_spec_bundle.py:83-92`; reana-client-go —
`pkg/specbundle/specbundle.go:133-144`, `pkg/specbundle/specbundle_test.go:138`

`_add_workflow_file_members` only walks a sub-tree when `workflow.file` lives in a
sub-directory; for a top-level workflow file it bundles the file alone:

```python
same_dir = os.path.realpath(workflow_dir) == os.path.realpath(base_dir)
if same_dir or not os.path.isdir(workflow_dir):
    _add_file_member(members, base_dir, workflow_file)
    return
```
(`reana_client/api/client.py:272-275`; the Go client mirrors this at
`specbundle.go:141-144`.)

So the very common Snakemake layout — a root-level `Snakefile` with
`include: "rules/common.smk"` — uploads only `Snakefile`, omitting
`rules/common.smk`. The same applies to a root-level `main.cwl` with
`$import: tools/tool.cwl`. The sandbox then cannot expand the DAG / resolve the
imports, so `reana-client validate` and `create` fail for a layout that worked
under the old client-side validation — while the sub-directory case
(`workflow/Snakefile`) bundles everything. The asymmetry is even asserted in
`tests/test_spec_bundle.py:83-92` (top-level file → only `{reana.yaml, Snakefile}`),
but no test covers a top-level file *with* an include, so the gap is invisible.

This was found independently by two reviewers, from opposite directions — a strong
signal it is real and will surprise users.

Suggested fix — fix the bundling rather than documenting a layout restriction:
for a top-level workflow file, still gather co-located includes (parse
`include:`/`$import` targets, or bundle sibling files referenced relatively). If a
restriction is unavoidable, surface it as an explicit client-side error naming the
missing referenced file, not a server-side load failure. Add a test for a
top-level workflow file with an include in both clients.

cmd/validate.go
```diff
@@ -0,0 +1,346 @@
...
+		displayer.DisplayMessage(warning.Message, displayer.Warning, true, out)
+	}
+	if o.pull {
+		for _, message := range checkImagesLocally(
+			report.Images, report.RuntimeUID, report.RuntimeGID,
+		) {
+			displayer.DisplayMessage(message, displayer.Warning, true, out)
+		}
+	}
+	if report.Valid {
```
tiborsimko, line 211: **🤖 SNDBX-13: Go client's `validate` failure path has no test coverage (LOW)**

*Component:* reana-client-go — `cmd/validate_test.go`, `cmd/validate.go:181-187,199-213`

`cmd/validate_test.go` exercises only the *success* path: every test server returns
`{"valid": true, ...}` (or checks flag forwarding / the `containsInt` helper). The
command's primary purpose — reporting an *invalid* spec — is never tested. The
whole block at `cmd/validate.go:199-213` (rendering `report.Errors` indented, then
returning `fmt.Errorf("%s is not a valid REANA specification", o.file)` for a
non-zero exit) is untested, as is the non-200 server-error branch at
`cmd/validate.go:181-187`. A regression that dropped the errors, returned `nil`
(exit 0 on an invalid spec), or mishandled a server error would pass CI green —
particularly relevant given the error-taxonomy issues in SNDBX-07. This is a test
gap rather than a present behaviour bug, hence LOW.

Suggested fix — add a test with a server returning
`{"valid": false, "errors": [{"message": "bad image"}]}` asserting both that the
error text appears in the output and that the command returns a non-nil error,
plus one asserting a non-200 response surfaces `report.Message`.

pkg/specbundle/specbundle.go
Sohn123: @tiborsimko I agree that the ideal situation would be to dynamically resolve imports etc. but I think that is just not possible, because to faithfully do that, we would need to run the engines where the client runs -> Exactly what we want to avoid. I think it is fair, that all files needed for successful validation need to be included as inputs. What do you think?

tiborsimko: I agree that we should not try to dynamically resolve Snakemake/CWL imports client-side, since that would pull us back toward running workflow-engine logic locally.

The current subdirectory rule already handles the common demo layout pretty well. For example, `reana-demo-cms-h4l` has:

```yaml
  inputs:
    parameters:
      input: workflow/input.yaml
    directories:
      - code
      - data
      - workflow
  workflow:
    type: snakemake
    file: workflow/Snakefile
```

Here `workflow.file` lives under `workflow/`, so the current bundle logic already includes the workflow subtree needed for validation, while excluding sibling `data/` and `code/`. That is the right behaviour: `data/` contains input ROOT files and should not be uploaded just for validation.

So I would not blindly upload all `inputs.files` / `inputs.directories`. They may contain large datasets, and `reana-client validate` should not unexpectedly upload those.

The remaining gap is narrower: top-level workflow files, e.g. `workflow.file: Snakefile`, that include/import other workflow-source files such as `rules/common.smk`.

For compatibility, I’d suggest a layered approach:

1. Keep the current behaviour: if `workflow.file` is in a subdirectory, bundle that workflow subtree. This already covers the common demo layout and avoids uploading sibling data directories such as `data/`.

2. Add an optional explicit field such as `workflow.files` or `workflow.includes` for users who want to declare extra workflow-definition files precisely.

3. For backward compatibility, in the top-level `workflow.file` case only, apply a conservative fallback:
   - include declared `inputs.files` entries that look like workflow source files, e.g. `Snakefile*`, `*.smk`, `*.cwl`, `*.tool`, maybe workflow YAML;
   - include declared `inputs.directories` only when the directory name looks workflow-related, e.g. `workflow`, `rules`, `cwl`, `snakemake`, not arbitrary directories like `data`;
   - never blindly bundle all `inputs.files` / `inputs.directories`.

All bundled paths should still pass the path-containment checks: relative only, no `..`, no absolute paths, no symlink escape, plus size/file-count caps.

This keeps existing common workflows working, gives users a clean explicit escape hatch, and avoids unexpectedly uploading huge input datasets during validation.

Talk IRL?

pkg/specbundle/specbundle.go
```diff
@@ -0,0 +1,290 @@
...
+	if specFilenames[member] {
+		return
+	}
+	members[member] = abs
+}
+
+// addWorkflowFileMembers adds workflow.file to the bundle, together with its
+// whole sub-tree when it lives in a sub-directory (to capture co-located
+// includes/imports/stage files). A top-level workflow file is added on its own
+// rather than gathering the entire specification directory.
```
michaelbuchar: This bundling rule looks too narrow for top-level workflows. When workflow.file lives at the bundle root, valid specs can still depend on sibling files, for example Snakemake include: "rules/common.smk" or CWL run: references next to the main workflow file. With the current behavior we only upload the root workflow file in that case, so validate/create regress on workflows that previously worked because the loader cannot find those sibling files in /validation/work/.... I reproduced this with a top-level Snakefile that includes rules/common.smk. Can we widen the bundle assembly here so root-level workflows also include the files they depend on, instead of treating “top-level” as “single-file”?

cmd/restart.go
```diff
@@ -125,7 +128,20 @@ func (o *restartOptions) run(cmd *cobra.Command) error {
...
+	if o.file != "" {
+		// Mirror the Python client: upload the modified specification into the
+		// shared restart workspace (the server re-loads and re-validates it at
+		// start) and pass it as `reana_specification` so the server reads the
+		// restart type from it.
+		reanaSpecification, err = readSpecificationFile(o.file)
+		if err != nil {
+			return err
+		}
+		if _, err = workflows.UploadFile(o.token, o.workflow, o.file); err != nil {
```
michaelbuchar: This restart path still uploads only the replacement `reana.yaml`, but after the server-side validation change the server reloads non-serial replacement specs from the workspace, not from the JSON payload alone. That means a restart whose new spec points to a different workflow file, or changes sibling files next to the workflow, can fail or silently reuse stale workspace contents because the rest of the raw bundle was never updated. I reproduced this with a Snakemake restart whose replacement spec referenced `Snakefile2`; `/start` failed because `Snakefile2` was missing from `/validation/work`. Can we upload the full replacement spec bundle here, using the same bundle-scoping logic as `create`/`validate`, instead of uploading only the YAML?

pkg/specbundle/specbundle.go
Sohn123: Fixed in latest version.

cmd/validate.go
tiborsimko: **🤖 Verified fixed** at `627fc98` — the Go client now tests both an invalid specification response and a server-error response through `TestValidateReportsInvalidSpecification` and `TestValidateSurfacesServerError`.

cmd/restart.go
tiborsimko: The restart `-f` semantics discussion has one remaining residual, tracked as [PR794-03 on reana-server#794](https://github.com/reanahub/reana-server/pull/794#discussion_r3590823593): a non-serial replacement uploaded under a non-canonical filename can be silently ignored in favour of the existing `reana.yaml`.

pkg/specbundle/specbundle.go
tiborsimko: The layered implementation (`workflow.files` plus the conservative top-level fallback) resolves the bundling design concern. Since `workflow.files` is now the explicit escape hatch users need for non-co-located sources, please add a user-facing example to the create/validate documentation; currently the mechanism is discoverable only through the schema.

pkg/specbundle/specbundle.go
tiborsimko: The layered `workflow.files` solution addresses this bundling issue. The remaining documentation request is recorded on the [main SNDBX-06 thread](https://github.com/reanahub/reana-client-go/pull/185#discussion_r3498964845).

cmd/restart.go
Sohn123: Ok we decided in an offline discussion the following:
- Restart means that all workflow files get reused only the reana.yaml file gets replaced if specified. 
- It is the user responsibility that all files necessary for validation are present in the workspace

pkg/specbundle/specbundle.go
```diff
@@ -0,0 +1,403 @@
...
+			continue
+		}
+		if err := addTreeMembers(members, baseDir, dirAbs); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Post sends the bundle to the given API path as multipart form data (each form
```
tiborsimko: **🤖 PR185-01: Bundle upload buffers the whole request and response in memory with no timeout (LOW)**

The Go client assembles the entire multipart body in memory before sending,
reads the whole response with an unbounded `io.ReadAll`, and uses an ad-hoc HTTP
client with no timeout:

```go
var body bytes.Buffer
writer := multipart.NewWriter(&body)
for relPath, localPath := range members {
    ...
    if _, err := io.Copy(part, file); err != nil { ... }
}
...
httpClient := &http.Client{
    Transport: &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
    },
}
resp, err := httpClient.Do(req)
...
respBody, err := io.ReadAll(resp.Body)
```

At the supported 100 MiB bundle size the client holds the whole bundle plus
multipart framing in memory, and then the full response body on top. The
`http.Client` has no `Timeout`, so a stalled or half-open connection hangs the
command indefinitely.

The Python client (reana-client #779) shares the same two problems —
`requests.post(files=...)` also builds the complete encoded body in memory and
passes no `timeout=` — so this is a paired concern; see PR779-03. Only the
unbounded response read is Go-specific.

Suggested fix — stream the multipart body through an `io.Pipe` (writing parts in
a goroutine) or a temporary file instead of a `bytes.Buffer`, bound the response
read (`io.LimitReader` with a sane maximum), and set an overall `Timeout` on the
`http.Client` (or a `context.WithTimeout` on the request).

Sohn123: Fixed. The bounded ZIP is written to a temporary file and passed to the generated go-swagger file parameter; its runtime streams multipart through an `io.Pipe`. The generated operation has a five-minute timeout, and typed response decoding replaces the ad-hoc unbounded `io.ReadAll` path.

client/client.go
```diff
@@ -15,13 +18,62 @@ import (
...
+}
+
+func (transport *boundedResponseTransport) RoundTrip(
+	request *http.Request,
+) (*http.Response, error) {
+	response, err := transport.transport.RoundTrip(request)
+	if err != nil {
+		return nil, err
+	}
+	if response.ContentLength > maxAPIResponseBytes {
```
tiborsimko, line 122: **🤖 PR185-02: Global response bound rejects downloads and other large API responses (MEDIUM)**

The PR185-01 fix installs the 16 MiB response limit on the shared transport used
by every generated API operation:

```go
const maxAPIResponseBytes = 16 * 1024 * 1024

func (transport *boundedResponseTransport) RoundTrip(
    request *http.Request,
) (*http.Response, error) {
    ...
    if response.ContentLength > maxAPIResponseBytes {
        _ = response.Body.Close()
        return nil, fmt.Errorf("%s", responseTooLargeText)
    }
    response.Body = &boundedResponseBody{
        body: response.Body, remaining: maxAPIResponseBytes,
    }
```

`pkg/workflows.DownloadFile()` obtains this same `ApiClient()` and passes a
`bytes.Buffer` to the generated download operation. Consequently any workflow
output or workspace archive larger than 16 MiB now fails with `REANA server
response exceeds 16 MiB limit`; a known `Content-Length` is rejected before the
download starts and an unknown/chunked response fails after 16 MiB. REANA
workspace downloads routinely exceed this size.

The blast radius is wider than downloads: every operation on the shared client
is capped, so sufficiently large log, status, file-listing, or disk-usage JSON
responses fail in the same way. The shared five-minute `http.Client.Timeout` is
an independent hard ceiling on every operation, including a legitimate download
over a slow connection even when its body is smaller than 16 MiB. Existing
tests prove that the global transport rejects an oversized response, but do not
prove that a large file transfer still succeeds.

Bounding JSON/control responses was appropriate for PR185-01, but doing it in a
global `RoundTripper` changes the semantics of existing APIs with potentially
large response bodies.

Suggested fix — apply the response-size bound only to generated operations that
decode small JSON/control payloads, especially validation/upload responses, and
leave file-transfer response bodies uncapped. Use an operation-specific client,
consumer, or transport policy rather than URL matching in the global transport.
Ideally make `DownloadFile()` stream to a caller-provided writer or destination
file instead of buffering the complete download in memory, with a separately
appropriate timeout policy. Add tests for both known-length and chunked
downloads larger than 16 MiB, while retaining the oversized-control-response
tests.

pkg/specbundle/specbundle.go
tiborsimko: **🤖 PR185-01 follow-up (2026-07-29): upload is fixed, but validation-bearing control calls still use the uncapped client (LOW)**

The original request-side problems are addressed at `ae3433a`: specification
archives are disk-backed and copied in bounded chunks, generated multipart
uploading streams from the temporary file, and create/validate use five-minute
operation timeouts. Validation also uses `ValidationAPIClient()`, whose transport
bounds control responses to 16 MiB.

The PR185-02 correction appropriately removed that bound and the global timeout
from the shared `ApiClient()` so downloads are not truncated or given an
inappropriate five-minute ceiling. However, creation still uses the uncapped
client, and the newly validation-bearing start/restart paths do so as well:

```go
// create.go
params := operations.NewCreateWorkflowParamsWithTimeout(5 * time.Minute)
api, err := client.ApiClient()

// start.go and restart.go
api, err := client.ApiClient()
startParams := operations.NewStartWorkflowParams()
```

Create has a per-operation timeout but no response bound. Start and restart have
neither the bounded transport nor a per-operation timeout, even though they now
run synchronous server-side validation and return `validation_warnings`. A
stalled connection can therefore hang those commands indefinitely, and all
three control responses can be decoded without the bound retained for explicit
validation.

Suggested fix — expose a bounded validation/control-response client and use it
for create, start, and restart as well as validate. Give start/restart explicit
operation deadlines long enough for the sandbox validation timeout, while
retaining the uncapped/no-global-deadline client for downloads and other
legitimately large operations. Add timeout and oversized-control-response tests
for the three commands.

Sohn123: Fixed. Create, validate, start, and restart now use the bounded control client and explicit five-minute operation deadlines. Their responses are capped at 16 MiB, while download operations continue using the uncapped client. Timeout and oversized-response behavior is tested for the affected commands.

client/client.go
Sohn123: Fixed. The shared client no longer has a global response cap or deadline, so large known-length and chunked downloads remain supported. The 16 MiB bound is isolated to validation-bearing control operations, with tests covering both large downloads and oversized control responses.

client/client.go
```diff
@@ -15,13 +18,78 @@ import (
...
 )
 
-// ApiClient provides a new API client used to communicate with the REANA server.
-func ApiClient() (*API, error) {
-	// disable certificate security checks
-	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{
-		InsecureSkipVerify: true,
+const (
+	apiRequestTimeout   = 5 * time.Minute
+	maxAPIResponseBytes = 16 * 1024 * 1024
```
tiborsimko, line 25: **🤖 PR185-04: Validation response cap conflicts with the server's valid response contract (LOW)**

The Go validation command uses `ControlAPIClient()`, which rejects response
bodies larger than 16 MiB. The validation endpoint accepts snapshots up to
100 MiB and returns the complete loaded `reana_specification` in its successful
report. A valid serial specification, or an expanded engine specification, can
therefore produce a response above 16 MiB: validation succeeds on the server
but the Go client reports `ErrResponseTooLarge`.

The CLI only consumes `valid`, errors, warnings, and optional environment
identities; it serialises and deserialises the generated payload but does not
use the returned specification. Raising the client cap would merely move the
contract mismatch and increase peak memory.

Suggested fix — preferably stop returning the full loaded specification from
the public validate response once server-side environment extraction is
complete. Otherwise define and publish a response-size limit that is
consistent with all accepted specifications and both clients. Add an
integration test with a valid response larger than 16 MiB.

pkg/specbundle/specbundle.go
```diff
@@ -0,0 +1,616 @@
...
+	info, err := os.Lstat(absoluteSpecification)
+	if err != nil {
+		return nil, err
+	}
+	if info.Mode()&os.ModeSymlink != 0 || !info.Mode().IsRegular() {
+		return nil, fmt.Errorf(
+			"the selected REANA specification must be a regular file",
+		)
+	}
+	contents, err := os.ReadFile(absoluteSpecification)
```
tiborsimko: **🤖 PR185-05: Specification parsing bypasses the configured bundle byte budget (LOW)**

`specbundle.Gather()` calls `os.ReadFile()` on the root specification before
the archive writer enforces `REANA_SPEC_BUNDLE_MAX_BYTES`. The restart path has
the same issue twice: `digestSpecificationFile()` reads the complete file to
hash it and `readRestartSpecification()` reads it again for lightweight YAML
parsing. A selected file can therefore exhaust client memory before the
documented 100 MiB budget has any effect.

This differs from the Python bundle-gathering implementation, which checks the
descriptor's size and performs a bounded read before YAML parsing. It is most
likely to be triggered by a mistaken local path, but it weakens the resource
contract and makes the Go client needlessly vulnerable to oversized local
inputs.

Suggested fix — introduce a shared descriptor-based bounded reader that checks
the file size and reads at most `bundleMaxBytes + 1`; use it for YAML parsing.
Hash restart replacements incrementally rather than with `os.ReadFile()`. Add
create, validate, and restart tests whose selected specification exceeds the
configured test budget.

cmd/validate.go
```diff
@@ -0,0 +1,361 @@
...
+	// Best-effort refresh; failures are ignored (a locally-built image may not
+	// exist in any registry). The run below is the authority -- it uses the local
+	// copy, auto-pulls a missing remote image, and fails clearly if absent.
+	_ = exec.CommandContext(ctx, cli, "pull", image).Run()
+	out, err := exec.CommandContext(
+		ctx,
+		cli,
+		"run",
+		"--rm",
+		"--entrypoint",
```
tiborsimko, line 274: **🤖 PR185-06: Local image inspection runs with network access and no resource limits (LOW)**

The `--pull` environment check runs `/bin/sh -c "id -u && id -G"` from the
inspected image with the container engine's default network, capability set,
PID allowance, writable filesystem, and CPU/memory limits. Overriding
ENTRYPOINT does not mean that image code is avoided: the shell, `id`
executable, dynamic loader, and libraries all come from the untrusted image.

Container namespaces, the default seccomp profile, and Docker/Podman's reduced
default capability set still provide meaningful isolation. However, a
malicious image can execute arbitrary image-controlled code with network access
to services reachable from the developer machine, or consume unbounded local
resources. `exec.CommandContext` kills the Docker/Podman client on timeout, but
does not guarantee that the daemon-managed container is stopped; `--rm` only
removes it after it exits. This is particularly surprising because the comment
explicitly claims that the image's own code does not run. The command is
opt-in and inspects an image the user intends to run on REANA, hence LOW rather
than MEDIUM.

Suggested fix — inspect a created-but-stopped container or exported filesystem
without executing image binaries. If execution is retained, add at least
`--network none`, a read-only root, dropped capabilities, no-new-privileges,
PID/CPU/memory limits, and a unique container/CID that is forcibly removed in
cleanup. Document the execution explicitly and test both the hardening flags
and timeout cleanup.

pkg/specbundle/specbundle.go
tiborsimko: **🤖 PR185-01 follow-up (2026-08-03): replacement upload still uses the timeout-zero client (LOW)**

The latest amendment correctly moves create, validate, start, and restart
control operations to `ControlAPIClient()` with explicit five-minute operation
deadlines. However, `restart -f` performs a separate replacement upload before
the bounded start call. `workflows.UploadFileAs()` constructs ordinary
`UploadFileParams` and creates `ApiClient()`, whose `http.Client.Timeout` is
zero and whose responses are uncapped.

That upload can still wait indefinitely on a stalled connection or response.
The current reply that create, validate, start, and restart are all bounded is
therefore accurate for their principal generated operations but not for the
complete `restart -f` command path.

Suggested fix — let `UploadFileAs()` accept the caller's API client and an
operation timeout, or add a bounded replacement-upload helper used by restart.
Keep ordinary downloads on the uncapped client. Add a command-level test in
which the replacement upload stalls, not only one where the final start call
stalls.

tiborsimko: **🤖 PR185-01 follow-up (2026-08-07): verified fully fixed at a5fc0e8**

The remaining `restart -f` upload now uses the already-created bounded control
client and `UploadControlFileAs()` constructs
`NewUploadFileParamsWithTimeout(controlOperationTimeout)`. The file remains
streamed through the generated multipart operation; the upload no longer falls
back to `ApiClient()` with a zero timeout or an uncapped response.

Together with the earlier temporary-file streaming and bounded control-client
changes, this closes the last residual from the original finding.

client/client.go
tiborsimko: **🤖 PR185-04 follow-up (2026-08-07): core contract fixed; shared report residual remains**

reana-server now removes the expanded `reana_specification` before returning a
successful public validation report. This fixes the concrete mismatch where a
valid specification close to the 100 MiB snapshot limit was reflected back to
the Go client and rejected by its 16 MiB control-response cap.

One narrower shared-server residual remains: parameter validation can still
emit one warning per unused parameter. In one fixture, 150,000 such parameters
turned a 3,788,975-byte compact input into a 16,950,039-byte report (about 4.5
times larger), so the Go client can still reject an otherwise successful
validation response. Raising the Go limit would only move that boundary; the
report needs a global entry or byte budget in reana-commons, as described in
[PR547-06](https://github.com/reanahub/reana-commons/pull/547#discussion_r3735082351).

The Go-side change requested by this finding is addressed. I would keep the
thread partial until the shared report itself is bounded.

pkg/specbundle/specbundle.go
tiborsimko: **🤖 PR185-05 follow-up (2026-08-07): verified fixed at a5fc0e8**

Specification loading now goes through a shared descriptor-based bounded
reader. It verifies the regular file, checks `Stat().Size()`, and reads through
`io.LimitReader(bundleMaxBytes + 1)` before YAML parsing. Restart digesting uses
`DigestSpecification()`, which hashes incrementally through the same limit
instead of calling `os.ReadFile()`.

The tests cover gather, direct read, and digest operations above a reduced byte
budget, as well as exact-byte hashing. The original memory and byte-contract
bypass is fully addressed.

client/client.go
Sohn123: The remaining shared-report residual is now fixed in reana-commons `8f46591`: every error/warning producer shares a 100-entry report budget, individual messages are bounded, and truncation is represented by one stable marker. Together with reana-server removing `reana_specification` from the public validation report, valid responses are no longer amplified by the expanded specification or unbounded parameter warnings. The Go control-response cap can remain unchanged, and this thread can be resolved.

cmd/validate.go
Sohn123: This is accepted intended behavior for now, consistently with the Python client. `--pull` is an explicit opt-in runtime compatibility probe for an image the user intends to execute on REANA; it deliberately runs the image-provided shell and `id` binaries to observe runtime UID/supplementary groups. I corrected the misleading comment that claimed no image code runs, but I am not changing the execution model or adding local sandbox flags in this PR. The residual local-engine risk is acknowledged and accepted.

pkg/specbundle/specbundle.go
```diff
@@ -0,0 +1,698 @@
...
+	}
+	file, err := openRegularFileBeneath(
+		filepath.Dir(absoluteSpecification),
+		filepath.Base(absoluteSpecification),
+		"REANA specification",
+	)
+	if err != nil {
+		return nil, err
+	}
+	metadata, err := file.Stat()
```
tiborsimko, line 504: **🤖 PR185-07: `Bounded streaming` relies only on a pre-upload stat (NIT)**

`OpenSpecification()` securely opens a regular file and rejects it when its
initial `Stat().Size()` exceeds `bundleMaxBytes`, but it then exposes the
unbounded `*os.File` to go-swagger's multipart transport. If another process
appends to the file between the stat and the upload read, the client transmits
past its configured byte ceiling.

The only actor able to grow this local file is on the user's own machine, and
the server independently caps the multipart request, so this is
defence-in-depth rather than an unbounded server exposure. The concrete issue
is that the "bounded specification for streaming" comment promises a property
implemented only as a point-in-time size check.

Suggested fix — expose a bounded `runtime.NamedReadCloser` (retaining the
`Name()` required by go-swagger) which cannot read beyond the accepted
length or `bundleMaxBytes + 1` and reports a clear error when growth crosses
the ceiling, or weaken the comment if stat-only enforcement is intentional.
Add a test which opens an under-limit file, appends to it, and asserts that the
multipart reader cannot stream beyond the configured budget.

client/client.go
tiborsimko: **🤖 PR185-04 follow-up (2026-08-10): environment output still exceeds the client cap**

The two previously discussed amplifiers are fixed: reana-server no longer
reflects the expanded specification, and reana-commons now applies one global
budget to its validation-report producers. The Go client does not need a
higher control-response limit for either case.

The complete endpoint response is not bounded yet, however. When
`environments=true`, reana-server appends the full distinct environment list
and one tag warning per image after the bounded report has been produced. A
valid fixture with 100,000 distinct `:latest` images generated a roughly
19.4 MB compact response, above this client's 16 MiB cap.

The remaining fix belongs in the server's optional environment
post-processing: it should share a count/byte budget with the report and emit a
stable truncation marker. I would therefore keep the cross-component response
contract open even though the Go-side implementation itself remains
appropriate.

Sohn123: The expanded specification is omitted from the public response, and optional environment identities/tag warnings now share the bounded validation-report budget and truncation marker. The Go client’s 16 MiB control-response cap can therefore remain unchanged.

pkg/specbundle/specbundle.go
Sohn123: The replacement stream now wraps the descriptor in a bounded NamedReadCloser, probes for overflow, and returns a deterministic size error without exposing bytes beyond the limit. Tests cover initial oversize, growth after opening, exact-size streaming, and closing.

cmd/validate.go
tiborsimko: **🤖 PR185-06 follow-up (2026-08-11): intended runtime probe documented; timeout cleanup remains (LOW remains)**

The amended comments now accurately state that `--pull` deliberately executes
the image-provided shell and `id` binaries. The explanation that this is an
explicit opt-in runtime compatibility probe, rather than static metadata
inspection, is a coherent product decision. The associated local-engine risk
is acknowledged, so this follow-up does not ask to change that execution model
in this PR.

One independent cleanup problem remains. `exec.CommandContext` kills the
Docker/Podman client when its context expires, but that does not guarantee that
the engine-managed container stops under the Docker daemon or Podman's
`conmon`. `--rm` removes it only after the container exits, and the command does
not assign a unique name or CID that can be removed after `CommandContext`
returns. An image whose shell never terminates can therefore leave a live
container on the developer's machine after the client reports the inspection
failure.

Suggested fix — assign each inspection a collision-resistant container name
or CID file and defer a bounded `docker`/`podman rm -f` cleanup for every exit
path, including context expiry. Limit cleanup to the identifier created by
this invocation, tolerate already-removed containers, and add a test that
cancels the context and verifies forced engine-side removal.

pkg/specbundle/specbundle.go
tiborsimko: **🤖 PR185-07 follow-up (2026-08-11): verified fixed at e2e047d**

Verified fixed. `OpenSpecification()` now returns a bounded reader that
enforces the byte limit while the multipart transport is consuming the file,
not only at open time. After exactly the allowed bytes it probes for one
overflow byte and returns a deterministic size error without exposing that
byte to the transport. Tests cover initial oversize, growth past the limit
while streaming, and exact-limit streaming with descriptor closure. The
“bounded streaming” contract is now enforced throughout the upload.

client/client.go
tiborsimko: **🤖 PR185-04 follow-up (2026-08-11): verified fixed at e2e047d**

Verified fixed across the server and shared-report changes. The public response
no longer reflects the expanded specification, every validation producer uses
the finite report budget, and optional environment identities/tag warnings now
consume that same bound. Truncation is represented by a stable marker. The Go
client's 16 MiB control-response cap can remain unchanged because the response
contract no longer contains either unbounded amplifier identified by this
thread.

cmd/validate.go
Sohn123: Fixed consistently with the Python client. Each inspection now uses a cryptographically random container name and defers a bounded docker/podman rm -f for that exact container on every exit path. The runtime probe model remains intentionally unchanged, and timeout cleanup is covered by a regression test.

pkg/workflows/manager.go
```diff
@@ -91,31 +95,83 @@ func GetWorkflowSpecification(
...
-	uploadParams.SetFile(string(fileData))
+	defer func() { _ = file.Close() }()
+	info, err := file.Stat()
+	if err != nil {
+		return "", fmt.Errorf("file %s could not be inspected: %w", localFileName, err)
+	}
 
-	api, err := client.ApiClient()
+	endpoint := serverURL.ResolveReference(&url.URL{
+		Path: fmt.Sprintf("/api/workflows/%s/workspace", url.PathEscape(workflow)),
```
tiborsimko: **🤖 PR185-08: Raw workspace uploads mishandle empty bodies and escaped workflow names (LOW)**

The hand-written request in `uploadFileAs()` introduces two wire-level regressions compared with the generated Swagger operation it replaces:

1. For a zero-byte file upload, `io.LimitReader(file, 0)` is still a non-nil request body. In Go's HTTP client, a non-nil body with `ContentLength == 0` has an unknown outgoing length, so the POST is sent chunked without a `Content-Length` header. The server now requires a declared upload length and rejects the request with 411.
2. `url.PathEscape(workflow)` is assigned to `URL.Path`. Serialising the URL escapes the percent signs again, so a legal name such as `analysis space` is sent as `analysis%2520space` rather than `analysis%20space`. The shared validation currently defines `REANA_WORKFLOW_NAME_ILLEGAL_CHARACTERS = ["."]`, so space, `#`, `?`, and `%` are permitted. The generated client on `master` encoded this path parameter once, making this a regression in this PR.

Suggested fix — use `http.NoBody` (or a nil body) for zero-byte uploads so Go emits `Content-Length: 0`; retain the limited reader for non-empty uploads. Put the decoded workflow name in `URL.Path` and let `URL.String()` escape it exactly once instead of assigning `PathEscape()` output to `Path`. A name containing `/` is a separate, pre-existing gap: creation permits it, but the Flask route cannot address it either way. An `httptest` regression should assert that an empty request has no chunked transfer encoding and that names containing space, `#`, `?`, or `%` appear exactly once-escaped in `RequestURI`.

Sohn123: Empty uploads now use http.NoBody, producing a known zero-length request without chunked transfer encoding. Workflow names are passed to URL.Path in decoded form so net/url escapes them exactly once. Tests cover empty bodies and names containing spaces, #, ?, and %.

pkg/specbundle/specbundle.go
```diff
@@ -0,0 +1,745 @@
...
+	Workflow struct {
+		Type        string   `yaml:"type"`
+		File        string   `yaml:"file"`
+		Files       []string `yaml:"files"`
+		Directories []string `yaml:"directories"`
+		Parameters  struct {
+			File string `yaml:"file"`
+		} `yaml:"parameters"`
+	} `yaml:"workflow"`
+	Inputs struct {
```
tiborsimko: **🤖 PR185-10: Go snapshots also require the legacy workflow-source fallback (MEDIUM)**

`specbundle.Gather()` mirrors the new explicit-only snapshot behavior: it adds `workflow.file`, `workflow.files`, and `workflow.directories`, while `specForBundle.Inputs` does not retain `files` or `directories`. Consequently, the Go client also omits auxiliary workflow definitions from existing specifications that historically declared their workflow directory under `inputs.directories`.

For example, the unchanged official `reana-demo-root6-roofit/reana-cwl.yaml` declares `workflow/cwl` under `inputs.directories`, and its primary `workflow.cwl` references `gendata.cwl` and `fitdata.cwl`. The resulting snapshot lacks both referenced files, so authoritative sandbox loading fails at `cwltool --pack` before the workflow can be created. Many current `reana-demo-*` specifications use the same legacy declaration pattern.

REANA analyses are designed to remain reusable and reproducible over time. Updated specifications should use the new explicit workflow-source fields, but an otherwise complete preserved analysis should not become unusable solely because it predates them.

Suggested fix — mirror the shared Python contract exactly: **when either `workflow.files` or `workflow.directories` is present, use the new explicit workflow scope; when both keys are absent, also include `inputs.files` and `inputs.directories` as the legacy validation scope.** Preserve key presence explicitly, for example with pointer-backed slices or the existing `yaml.Node`, rather than checking only whether a decoded slice is non-empty, so an explicitly empty list selects the new behavior.

Uploading the complete legacy input scope may be inefficient, but only compatibility mode pays that cost; the existing path-safety and resource limits can remain unchanged. This parity is necessary because reana-server re-gathers the uploaded snapshot with the shared Python helper and requires exact equality with the archive member names.

`addDirectory()` has the same overlap issue as the Python helper: it infers emptiness from the delta in `members`, so a legacy input directory containing the already selected `workflow.file` fails. Count regular files encountered during traversal instead of newly added members, matching the shared fix.

Please test the unchanged ROOT6 CWL layout and absent, explicitly empty, and non-empty workflow-source declarations. The selected members and errors should match the Python client.

cmd/create.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer func() { _ = bundle.Close() }()
+
+	params := operations.NewCreateWorkflowParamsWithTimeout(
+		controlOperationTimeout,
+	)
+	params.SetAccessToken(&o.token)
+	params.SetBundle(bundle)
```
tiborsimko, line 159: **🤖 PR185-12: generated multipart bundle uploads are sent chunked and are lost by the deployed server (HIGH)**

Both new bundle uploads — `params.SetBundle(bundle)` in `cmd/create.go` and in `cmd/validate.go` — go through go-swagger's generated file parameter. In `go-openapi/runtime` v0.32.6, `writeMultipartBody()` always assembles a multipart body through an `io.Pipe`:

```go
func (r *Request) writeMultipartBody(ctx context.Context, mediaType string) io.Reader {
	pr, pw := io.Pipe()
	mp := multipart.NewWriter(pw)
	...
	go r.streamMultipartParts(ctx, mp, pw)
	return pr
}
```

`http.NewRequestWithContext` cannot infer a length from a pipe, so `ContentLength` stays unset and Go frames the request with `Transfer-Encoding: chunked`. Capturing both operations against a local server confirms it:

```text
POST /api/workflows/validate   Content-Length=absent  Transfer-Encoding=[chunked]
POST /api/workflows            Content-Length=absent  Transfer-Encoding=[chunked]
```

The multipart payload itself is well formed — a standards-compliant Go server parses one `bundle` part from it. The failure is specific to how reana-server is deployed. The chart runs uWSGI's HTTP router with post-buffering and without `http-chunked-input`:

```ini
http = 0.0.0.0:5000
buffer-size =  10240
post-buffering = true
```

uWSGI documents post-buffering as incompatible with chunked input, and receiving chunked bodies through the HTTP router requires `--http-chunked-input`. The body is therefore not delivered and Flask sees an empty `request.files`.

Against a deployment of this PR set, with a binary verified as built from the current branch head, `create` returns the message the create endpoint emits only on that empty-`request.files` branch:

```
A workflow specification bundle must be uploaded.
```

`validate` fails on the same deployment, but surfaces only the generated client's generic error rather than the server's own `No specification bundle files were provided.` body, because the validation operation's 400 response discards the payload. So every Go `create`, `validate`, and the new `run` command are broken on a standard REANA installation, and one of the two gives no usable diagnostic.

The Python client is unaffected, which is why this went unnoticed: `requests` builds the multipart body in memory and sends `Content-Length` (measured at 333 bytes for an equivalent bundle), never chunked.

Two test gaps let this through. On the Go side, `cmd/create_test.go` asserts only that the request's content type starts with `multipart/form-data`; it never parses the body or inspects framing, so it passes while the wire format is unusable. On the server side, `tests/test_views.py` builds its chunked-upload cases by setting `environ["wsgi.input_terminated"] = True`, which models a WSGI frontend that the chart's uWSGI configuration cannot actually provide.

Suggested fix — send these two requests with a known `Content-Length`. Please avoid buffering the body in memory: the bundle limit is 100 MiB, so materialising a multipart body in RAM would trade this bug for a client-side memory problem. Two workable shapes are a length-known multipart reader, which computes the exact encoded length from the part headers, boundary and the underlying temporary archive's size, and streams the parts without holding them, or a multipart body spooled to a temporary file whose size is then used as the length. Either keeps the generated operations untouched if applied at the transport layer in `client/client.go`, alongside the existing `boundedResponseTransport`.

Please add a Go regression test that asserts a positive `Content-Length`, absence of chunked transfer encoding, and a `bundle` part that parses back to the archive; the current `cmd/create_test.go` check passes on a request the deployed server cannot read. Note that `_cap_bundle_request_body()` in reana-server should stay as it is — bounding a chunked upload remains a valid application-layer defence under any frontend that can deliver one.

pkg/specbundle/specbundle.go
tiborsimko: **🤖 PR185-10 follow-up (2026-08-13): the Go fallback matches Python, except for key presence inherited through a merge key (MEDIUM)**

Thanks, the Go client now implements the agreed rule and matches the Python gatherer on the cases that matter. Verified against `specbundle.Gather()` and `gather_validation_members()` on identical trees: a legacy specification with `workflow.file` and no explicit scope keys selects the inputs; an explicit `workflow.files: []` suppresses the fallback; an inline serial specification never pulls inputs into the validation scope. The `addDirectory()` emptiness correction is also present and equivalent.

One narrow divergence remains. `mappingValue()` scans the syntactic children of the raw `workflow` node, while `specification.Workflow` is read from the decoder, which resolves aliases and `<<` merge keys. When `files` or `directories` arrives through a merge, Go sees no explicit scope and enters compatibility mode, whereas PyYAML resolves the merge and takes the narrow scope. The server then rejects the Go client's archive as not matching its declared validation scope, while the Python client succeeds:

```yaml
inputs:
  files: [data.txt]
workflow:
  <<: &defaults
    files: []
  type: cwl
  file: workflow/main.cwl
```

Go selects `data.txt`, `reana.yaml`, `workflow/main.cwl`; Python selects `reana.yaml`, `workflow/main.cwl`.

The anchor is deliberately defined inline and used once. That detail matters: a merge base that is referenced twice — for example an anchored top-level mapping — makes the merged list the *same* object in both places, so `validate_json_tree()` already rejects the specification on the server, and a top-level anchor key is separately rejected by the schema's `additionalProperties: false`. Those forms fail on both clients, just with different messages, so the genuine residual is only the single-use inline form above.

Suggested fix — make presence detection use the same resolution the values use, for example by declaring the two scope fields as pointer-backed slices (`Files *[]string`, `Directories *[]string`) so a nil pointer means absent and a non-nil empty slice means present-but-empty. That removes the raw-node scan entirely and keeps the two reads consistent. Please add one regression test covering a scope key inherited through a merge key. Given the narrow trigger this is not merge-blocking.

tiborsimko: **🤖 PR185-10 follow-up (2026-08-14): verified fixed (MEDIUM)**

Verified on current head `93a809d` against the Python shared gatherer. A
nine-case parity harness now selects identical members for legacy fallback,
explicit empty scope, inline serial workflows, overlapping directories, a
single-use inline merge key, and both allowed and rejected canonical input
declarations. Presence and values are therefore read through the same resolved
representation, including the previously divergent merge-key case. This fully
addresses the finding.

cmd/create.go
tiborsimko: **🤖 PR185-12 follow-up (2026-08-14): verified fixed (HIGH)**

Verified on current head `93a809d` through a real `ControlAPIClient()` request
against an HTTP test server. The multipart request carries an exact
`Content-Length` of 411 bytes, has no chunked `Transfer-Encoding`, and parses
back to one intact bundle file. The same request path also returns the server's
structured 400 payload to the caller. This fully addresses the unusable request
framing reported here.

tiborsimko (approved):

